PR #4, opened 2020-11-21
Add Rails type definitions

pocke:
This pull request adds RBS files for Ruby on Rails. They are imported from pocke/rbs_rails.

# Added RBS files

I added three sorts of files. 

First, `-generated` suffixed files, such as `activerecord-generated.rbs`.
They are generated by `generate.rb`, which is based on `rbs prototype rb`, and not modified by hand.

Seocnd, `patch.rbs`.
They include monkey patches to fix errors. I designed them as a place to put RBSs that are temporarily necessary due to bug of RBS. 

Last, other rbs files, such as `activerecord.rbs`.
They have hand-written RBS files. If we want to improve the generated types, we can write types to the files (not `-generated` files).

# Directory structure

Ruby on Rails has several gems, such as activereocrd, actionpack, etc.. But they need the same script to generate/test RBSs for them.
So I put the scripts under Railties directory. Other gems don't have any scripts/tests.

The Railties directory has pocke/rbs_rails as a git submodule. pocke/rbs_rails is necessary to generate RBSs and test them.
ref: https://github.com/pocke/rbs_rails/pull/65 The submoduled rbs_rails refers the commit in this pull request.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,9 @@
 [submodule "gems/rainbow/3.0/_src"]
 	path = gems/rainbow/3.0/_src
 	url = https://github.com/sickill/rainbow.git
+[submodule "gems/railties/6.0.3.2/_src"]
+	path = gems/railties/6.0.3.2/_src
+	url = https://github.com/rails/rails.git
+[submodule "gems/railties/6.0.3.2/_rbs_rails"]
+	path = gems/railties/6.0.3.2/_rbs_rails
+	url = https://github.com/pocke/rbs_rails.git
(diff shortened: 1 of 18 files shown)


pocke:
I updated rbs gem to v0.19 for generating Rails's RBS files. https://github.com/ruby/gem_rbs/pull/4/commits/62aadab7c8fdd3d4dbd5cd8de26fde612f47082c
This change is necessary to use RBS gem's new features, such as `extend self` by `rbs prototype rb`. 


I moved interfaces of action controller to under the namespace. https://github.com/ruby/gem_rbs/pull/4/commits/0db811a7a352f03718969eb2f84abc26098aaea4

PR #6, opened 2020-12-17
[retryable] Fix `sleep` and `sleep_method` signatures

ybiquitous:
- Fix the incorrect return type (`void` -> `Numeric`) of `sleep` lambda.
- Change `real` to `Numeric` because `Kernel.sleep` receives `Numeric`.
  (See <https://github.com/ruby/rbs/blob/09f21280f6d8041b3d71473b0551199a68818e92/core/kernel.rbs#L374-L375>)

See also <https://github.com/nfedyashev/retryable/blob/aa51cbd7ea570e223161571fe97eb8ca7ce8a535/lib/retryable.rb#L84-L85>.


soutaro (approved):
😪

ybiquitous:
Thank you!

PR #7, opened 2020-12-17
[rainbow] Support X11 color name methods

ybiquitous:
This change adds X11 color name methods that Rainbow supports like `Rainbow('Hi').aliceblue`.

See also:
- https://github.com/sickill/rainbow#x11-colors
- https://github.com/sickill/rainbow/blob/v3.1.0/lib/rainbow/x11_color_names.rb
- https://github.com/sickill/rainbow/blob/v3.1.0/lib/rainbow/presenter.rb#L126-L132

---

It is a hard task to add these methods manually, so this change uses a script to generate code.
For example:

```console
$ ruby _scripts/gen_x11_colors.rb sig
```


soutaro (approved):
😲

PR #15, opened 2021-02-13
Refactor for `find` and `ids` methods

soutaro:
This is for https://github.com/pocke/rbs_rails/pull/105.

Adding `PrimaryKey` type parameter allows supporting AR models with non-integer primary key.


pocke (approved):
Thanks!

PR #17, opened 2021-02-17
Fix module-self-type requirements

soutaro:
Implementing validation on _module-self-type_ requirements uncovered some problems in Rails RBS.

### DIFF
--- a/gems/actionview/6.0.3.2/actionview-generated.rbs
+++ b/gems/actionview/6.0.3.2/actionview-generated.rbs
@@ -9110,6 +9110,8 @@ module ActionView
 
     def find_all_with_query: (untyped query) -> (untyped | ::Array[untyped])
 
+    def each: () { (untyped) -> void } -> untyped
+
     private
 
     def _find_all: (untyped path, untyped prefixes, untyped args) -> (untyped | ::Array[untyped])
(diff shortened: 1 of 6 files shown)


gems/actionview/6.0.3.2/actionview-generated.rbs
```diff
@@ -9110,6 +9110,8 @@ module ActionView
 
     def find_all_with_query: (untyped query) -> (untyped | ::Array[untyped])
 
+    def each: () { (untyped) -> void } -> untyped
```
soutaro, line 9113: `#each` was missing because it is delegated.

https://github.com/rails/rails/blob/6-0-stable/actionview/lib/action_view/path_set.rb#L16

gems/activemodel/6.0.3.2/activemodel-generated.rbs
```diff
@@ -1441,7 +1441,7 @@ module ActiveModel
     #     # Will yield :name and "can't be blank"
     #     # then yield :name and "must be specified"
     #   end
-    def each: () { (untyped, untyped) -> untyped } -> untyped
+    def each: () { ([untyped, untyped]) -> untyped } -> untyped
```
soutaro, line 1444: Rails 6.0 accepts a pair. (6.1 accepts only one argument.)

gems/activerecord/6.0.3.2/activerecord-generated.rbs
```diff
@@ -18051,6 +18051,8 @@ module ActiveRecord
       def relation_class_for: (untyped klass) -> untyped
     end
 
+    def each: () { (untyped) -> void } -> untyped
```
soutaro, line 18054: `#each` was missing because it is delegated.

https://github.com/rails/rails/blob/6-0-stable/activerecord/lib/active_record/relation/delegation.rb#L88

pocke (approved):
Thanks!

PR #21, opened 2021-03-04
Add RBS for Parallel gem

ybiquitous:
https://github.com/grosser/parallel/tree/v1.20.1

Run the commands for testing:

```shell
cd gems/parallel/1.20
../../../bin/run _scripts/test --repo=../../..
```

### DIFF
--- a/gems/parallel/1.20/parallel.rbs
+++ b/gems/parallel/1.20/parallel.rbs
@@ -1,0 +1,83 @@
+module Parallel
+  class Break < StandardError
+    attr_reader value: untyped
+
+    def initialize: (?untyped value) -> void
+  end
+
+  class Kill < Break
+  end
+
+  Stop: Object
+
+  VERSION: String
+
+  Version: String
+
+  def self.all?: [T] (Enumerable[T] | (^() -> T | untyped) array, # TODO: `T | Stop` should be better than `T | untyped`.
+                      ?in_processes: Integer,
+                      ?in_threads: Integer,
+                      ?progress: _ToStr,
+                      ?start: ^(T item, Integer index) -> untyped,
+                      ?finish: ^(T item, Integer index, untyped result) -> untyped) { (T) -> boolish } -> bool
+
+  def self.any?: [T] (Enumerable[T] | (^() -> T | untyped) array,
+                      ?in_processes: Integer,
+                      ?in_threads: Integer,
+                      ?progress: _ToStr,
+                      ?start: ^(T item, Integer index) -> untyped,
+                      ?finish: ^(T item, Integer index, untyped result) -> untyped) { (T) -> boolish } -> bool
+
+  def self.each: [T, U] (Enumerable[T] array,
+                         ?in_processes: Integer,
+                         ?in_threads: Integer,
+                         ?progress: _ToStr,
+                         ?start: ^(T item, Integer index) -> untyped,
+                         ?finish: ^(T item, Integer index, U result) -> untyped) { (T) -> U } -> Enumerable[U]
+               | [T, U] ((^() -> T | untyped) array,
+                         ?in_processes: Integer,
+                         ?in_threads: Integer,
+                         ?progress: _ToStr,
+                         ?start: ^(T item, Integer index) -> untyped,
+                         ?finish: ^(T item, Integer index, U result) -> untyped) { (T) -> U } -> (^() -> T | untyped)
+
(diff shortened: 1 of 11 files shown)


gems/parallel/1.20/parallel.rbs
```diff
@@ -0,0 +1,83 @@
...
+  class Kill < Break
+  end
+
+  Stop: Object
+
+  VERSION: String
+
+  Version: String
+
+  def self.all?: [T] (Enumerable[T] | (^() -> T | untyped) array, # TODO: `T | Stop` should be better than `T | untyped`.
```
ybiquitous, line 17: [question] When using `Stop` instead of `untyped`, 

```diff
--- a/gems/parallel/1.20/parallel.rbs
+++ b/gems/parallel/1.20/parallel.rbs
@@ -14,7 +14,7 @@ module Parallel
 
   Version: String
 
-  def self.all?: [T] (Enumerable[T] | (^() -> T | untyped) array, # TODO: `T | Stop` should be better than `T | untyped`.
+  def self.all?: [T] (Enumerable[T] | (^() -> T | Stop) array, # TODO: `T | Stop` should be better than `T | untyped`.
                       ?in_processes: Integer,
                       ?in_threads: Integer,
                       ?progress: _ToStr,
```

then the following error occurs. How can I fix it? 🤔 

```console
$ ../../../bin/run _scripts/test --repo=../../..
...
gems/parallel/1.20/parallel.rbs:17:50...17:54: Could not find ::Stop (RBS::NoTypeFoundError)
...
```

gems/parallel/1.20/parallel.rbs
```diff
@@ -0,0 +1,83 @@
...
+                                   ?in_threads: Integer,
+                                   ?progress: _ToStr,
+                                   ?start: ^(T item, Integer index) -> untyped,
+                                   ?finish: ^(T item, Integer index, U result) -> untyped) { (T, Integer) -> U } -> Array[U]
+
+  def self.physical_processor_count: () -> Integer
+
+  def self.processor_count: () -> Integer
+
+  def self.worker_number: () -> Integer
```
ybiquitous, line 82: [note] When evaluating `Parallel.worker_number` without a block, it returns `nil`:

```
irb(main):003:0> Parallel.worker_number
=> nil
```

But this usage is incorrect and we should not want nil-check for `Parallel.worker_number`, so I decide to return `Integer` instead of `Integer?`.

If the return type would be `Integer?`, the following code would encounter a type error:

```
number.rb:1:63: [error] Type `(::Integer | nil)` does not have method `/`
│
└ Parallel.each(1..5) { |i| puts "#{i}: #{Parallel.worker_number / 2.0}" }
                                                                 ~
```

gems/parallel/1.20/parallel.rbs
```diff
@@ -0,0 +1,83 @@
...
+                                    ?in_threads: Integer,
+                                    ?progress: _ToStr,
+                                    ?start: ^(T item, Integer index) -> untyped,
+                                    ?finish: ^(T item, Integer index, U result) -> untyped) { (T, Integer) -> U } -> Enumerable[T]
+                          | [T, U] ((^() -> T | untyped) array,
+                                    ?in_processes: Integer,
+                                    ?in_threads: Integer,
+                                    ?progress: _ToStr,
+                                    ?start: ^(T item, Integer index) -> untyped,
+                                    ?finish: ^(T item, Integer index, U result) -> untyped) { (T, Integer) -> U } -> (^() -> T | untyped)
```
ybiquitous, line 55: [note] This overloading is too ugly, but I cannot find a good way... 😢

gems/parallel/1.20/parallel.rbs
pocke: RBS doesn't treat constants as types. So you cannot access `Stop` as a type in RBS.
I guess it will be useful if we can put a constant as a type, so could you open an issue to ruby/rbs repo? cc/ @soutaro 

By the way, we can improve the type by using a literal type instead.


```rbs
class Parallel
  type stop = :__type_of_parallel_stop__
  Stop: stop
end
```

it solves the problem, but it is incorrect. So personally I think untyped is better.

ybiquitous: @pocke Thanks for your kind feedback! I've opened an issue: ruby/rbs#623. 👍 

Could you please provide further information if necessary?

PR #29, opened 2021-05-19
Add block signature to `_ActiveRecord_Relation#find`

alts:
When `find` is called with a block, ActiveRecord relations defer
to the underlying `Enumerable` implementation.

### DIFF
--- a/gems/activerecord/6.0.3.2/activerecord.rbs
+++ b/gems/activerecord/6.0.3.2/activerecord.rbs
@@ -84,6 +84,7 @@ interface _ActiveRecord_Relation[Model, PrimaryKey]
   def find: (PrimaryKey id) -> Model
           | (Array[PrimaryKey]) -> Array[Model]
           | (*PrimaryKey) -> Array[Model]
+          | () { (Model) -> boolish } -> Model
   def first: () -> Model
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Looks good to me. 👏

gems/activerecord/6.0.3.2/activerecord.rbs
```diff
@@ -84,6 +84,7 @@ interface _ActiveRecord_Relation[Model, PrimaryKey]
   def find: (PrimaryKey id) -> Model
           | (Array[PrimaryKey]) -> Array[Model]
           | (*PrimaryKey) -> Array[Model]
+          | () { (Model) -> boolish } -> Model
```
pocke, line 87: `find` is nil-able. For example: `User.all.find{false} # => nil`

```suggestion
          | () { (Model) -> boolish } -> Model?
```

soutaro: Ah. Right... 🙇‍♂️

alts: Ah, thank you!

pocke (approved):
Thank you!

PR #30, opened 2021-05-26
Fix AR where opts type.

floor114:
According to the docs, you could pass String, Array, Hash, or nothing (for `where.not`) as an argument.


pocke:
Thanks!

PR #33, opened 2021-06-29
Fix optional block in ActiveRecord::Base initialize

jocmp:
Annotates the block in ActiveRecord::Base#initialize to provide type parity with the underlying library since the block is optional.

### Example of issue
```
# activerecord-generated.rb:4:12: [error] The method cannot be called without a block
# │ Diagnostic ID: Ruby::RequiredBlockMissing
# │
# └ user = User.new
#               ~~~

# Detected 1 problem from 1 file
```

### DIFF
--- a/gems/activerecord/6.0.3.2/_test/Steepfile
+++ b/gems/activerecord/6.0.3.2/_test/Steepfile
@@ -1,0 +1,25 @@
+target :test do
+  signature "."
+  check "."
+
+  repo_path "../../../"
+
+  library "pathname"
+  library "logger"
+  library "mutex_m"
+  library "date"
+  library "monitor"
+  library "singleton"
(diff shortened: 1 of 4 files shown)


gems/activerecord/6.0.3.2/_test/Steepfile
```diff
@@ -0,0 +1,25 @@
+target :test do
+  signature "."
+  check "."
+
+  repo_path "../../../"
+
+  library "pathname"
```
jocmp, line 7: Required libraries. This is the same list as the one found on the [pock/rbs_rails](https://github.com/pocke/rbs_rails#steep-integration) README.

soutaro (approved):
Thanks!

PR #36, opened 2021-07-12
Add stub signatures for Rack gem

mame:
I created the prototype of signatures by applying TypeProf to Rack and
editting the result manually.

```
bundle exec typeprof -Ilib example/lobster.ru -o sig/rack.rbs
```

I've checked that `rbs validate` passed.
There are still many glitches, but I believe that this is a good start
to create the RBS of Rack.

### DIFF
--- a/gems/rack/2.2.2/_test/Steepfile
+++ b/gems/rack/2.2.2/_test/Steepfile
@@ -1,0 +1,6 @@
+target :test do
+  check "."
+
+  repo_path "../../../"
+  library "rainbow"
+end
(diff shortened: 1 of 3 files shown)


gems/rack/2.2.2/_test/Steepfile
```diff
@@ -0,0 +1,6 @@
+target :test do
+  check "."
+
+  repo_path "../../../"
+  library "rainbow"
```
pocke, line 5: why rainbow?

pocke (approved):
LGTM :+1:

PR #46, opened 2021-07-23
Syntax check for RBS with GitHub Actions

pocke:
This PR enables GitHub Actions to syntax check RBS in this repository.

Syntax check is not enough to ensure that RBS is valid semantically, but it is better than nothing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,19 @@
+name: Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request: {}
+
+jobs:
+  syntax_check:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pocke:
The YAML looks valid but GHA doesn't run, could you enable GHA from the repository setting? @mame @soutaro

mame:
It looks like GHA is already enabled. Maybe we need to merge the YAML to run it?

pocke:
Thank you. I'll merge it to try starting GHA :rocket:

pocke:
https://github.com/ruby/gem_rbs_collection/runs/3141278274 GHA successfully ran! The merge is the trigger. Thanks for your advice!

PR #50, opened 2021-08-10
Fix Redis#initialize parameter name

cky:
Currently, `Redis#initialize`'s signature has two `path` parameters. This breaks tools like RubyMine, which (rightly) expects no duplicates in parameter names. Looking at `Redis::Client`'s options, it appears that the parameter preceding `host` is `scheme`, which is absent in the current RBS file.


pocke (approved):
Thanks!

PR #59, opened 2021-09-06
Update httpclient type

ksss:
Some of the patterns were missing and will be corrected.
Strictly, the `| String` part should be `_ToS`, but since most objects have the `to_s` method.
It may induce unintended use.
so it is `| String`.


pocke:
> Strictly, the `| String` part should be `_ToS`, but since most objects have the `to_s` method.
> It may induce unintended use.
> so it is `| String`.

I'm not sure rejecting an object with `to_s` is the best way. But I think this patch is acceptable because the end-users can define an overload if they need.

PR #60, opened 2021-09-10
Extend ClassMethods for ActiveRecord::Base

ksss:
`extend ActiveSupport::Concern` will automatically extend the `ClassMethods`module.

RBS cannot follow this metaprogramming, so the following code adds the actual ClassMethods that are extended.

```rb
puts ActiveRecord::Base.singleton_class.ancestors.map(&:to_s).select { |s|
  /ClassMethods$/.match?(s)
}.map{ |s|
  "extend ::#{s}"
}.sort
```

### DIFF
--- a/gems/activerecord/6.0/activerecord-generated.rbs
+++ b/gems/activerecord/6.0/activerecord-generated.rbs
@@ -13766,7 +13766,7 @@ module ActiveRecord
       # Determines if one of the attributes passed in is the inheritance column,
       # and if the inheritance column is attr accessible, it initializes an
       # instance of the given subclass instead of the base class.
-      def new: (?untyped? attributes) { () -> untyped } -> untyped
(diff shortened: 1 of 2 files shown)


gems/activerecord/6.0/activerecord-generated.rbs
```diff
@@ -13766,7 +13766,7 @@ module ActiveRecord
       # Determines if one of the attributes passed in is the inheritance column,
       # and if the inheritance column is attr accessible, it initializes an
       # instance of the given subclass instead of the base class.
-      def new: (?untyped? attributes) { () -> untyped } -> untyped
+      def new: (?untyped? attributes) ?{ () -> untyped } -> untyped
```
ksss, line 13769: This is a fix to make test pass.

pocke (approved):
Thanks!

PR #61, opened 2021-09-12
Add activestorage signatures

ksss:
Added signature for activestorage v6.0.
APIs that are likely to be used in practice have been added.
There are no changes on the activerecord side, and `ActiveRecord::Base` is extended only when the signature of activestorage is read.

The reason why I changed the way to divide the files into rb files is because I consider it easier to find the files and to investigate the differences between versions.


gems/activestorage/6.0/app/models/active_storage/blob.rbs
```diff
@@ -0,0 +1,150 @@
...
+  # data might get overwritten on the service.
+  def upload: (untyped io, ?identify: bool identify) -> untyped
+
+  def unfurl: (untyped io, ?identify: bool identify) -> untyped
+
+  def upload_without_unfurling: (untyped io) -> untyped
+
+  # Downloads the file associated with this blob. If no block is given, the entire file is read into memory and returned.
+  # That'll use a lot of RAM for very large files. If a block is given, then the download is streamed and yielded in chunks.
+  def download: () { () -> untyped } -> untyped
```
pocke, line 120:
```suggestion
  def download: () { (String) -> void } -> void
              | () -> String
```

We can improve `download` signature like the above.

* `download` allows calling without block and it returns a String.
* `downlaod` yields the block with a String.

gems/activestorage/6.0/app/models/active_storage/blob.rbs
```diff
@@ -0,0 +1,150 @@
...
+  # By default, the tempfile is created in <tt>Dir.tmpdir</tt>. Pass +tmpdir:+ to create it in a different directory:
+  #
+  #   blob.open(tmpdir: "/path/to/tmp") do |file|
+  #     # ...
+  #   end
+  #
+  # The tempfile is automatically closed and unlinked after the given block is executed.
+  #
+  # Raises ActiveStorage::IntegrityError if the downloaded data does not match the blob's checksum.
+  def open: (?tmpdir: untyped? tmpdir) { () -> untyped } -> untyped
```
pocke, line 135: `open` yields the block with a tempfile.

```suggestion
  def open: (?tmpdir: untyped? tmpdir) { (Tempfile) -> untyped } -> untyped
```


https://github.com/rails/rails/blob/0d304eae601f085274b2e2c04316e025b443da62/activestorage/app/models/active_storage/blob.rb#L216-L219
https://github.com/rails/rails/blob/0d304eae601f085274b2e2c04316e025b443da62/activestorage/lib/active_storage/service.rb#L86-L88
https://github.com/rails/rails/blob/0d304eae601f085274b2e2c04316e025b443da62/activestorage/lib/active_storage/downloader.rb#L11-L17

gems/activestorage/6.0/app/models/active_storage/blob.rbs
ksss: Thank you!

gems/activestorage/6.0/app/models/active_storage/blob.rbs
ksss: ✨

ksss:
Thank you for reviewing.
I applied your suggestion code.

pocke (approved):
Thanks!

PR #62, opened 2021-09-16
Add graphql gem

ksss:
I wrote the rbs for the [graphql](https://github.com/rmosolgo/graphql-ruby) gem.

I would like to use rbs for coding that uses the graphql gem, but it is noisy with many errors.

The rbs to be added in this PR aims to reduce this noise errors as much as possible and focus on the application part.

The graphql gem is a lot of code, but I extracted the code that is actually used, focusing on the generated code.


pocke (approved):
Thanks! It is a great first step :chart_with_upwards_trend:

ksss:
👣

PR #63, opened 2021-09-18
Fix activestorage v6.0

ksss:
# Add dependency to tempfile

activestorage needs dependency to `Tempfile`.
Without this library, rbs will not be able to be loaded.

# Should use v6.0 declaration

The declaration of `has_one_attached` and `has_many_attached` was used v6.1.


pocke (approved):
Thanks!

PR #66, opened 2021-09-23
add default_timezone= fix #65

### LINKED ISSUE #65: can't use ActiveRecord::Base.default_timezone
How to define `default_timezone` ?

```ruby
ActiveRecord::Base.default_timezone = :local
```
this gets an error below;

```
Type `singleton(::ActiveRecord::Base)` does not have method `default_timezone=`(Ruby::NoMethod)
```

POPPIN-FUMI:


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -39,6 +39,7 @@ module ActiveRecord
     extend ::ActiveSupport::Callbacks::ClassMethods
 
     def self.abstract_class=: (bool) -> void
+    def self.default_timezone=: (*untyped) -> untyped
```
pocke, line 42:
```suggestion
    def self.default_timezone=: (untyped) -> untyped
```


This method doesn't receive a variadic argument but it receives just one argument.

POPPIN-FUMI: Yea thank you!

pocke (approved):
Thanks! :rocket:

PR #76, opened 2021-10-30
Add aws-sdk-s3 gem

ksss:
This PR will be able to check type like the following code.

```rb
resp = Aws::S3::Client.new.get_object(bucket: 'test', key: 'test')
resp.body.read
````

### aws_client_types_generator.rb

Automatically generates type definitions from json files in the [aws-sdk-ruby](https://github.com/aws/aws-sdk-ruby).
`2006-03-01_api-2.rbs` can be generated with just execute `$ bundle exec rake`.
This script should be able to handle clients other than S3 in effect, but I'm using aws-sdk-s3 as the scope for now.

### Aws::S3::Client

- Only use keyword argument.
    - It can set the required keyword and option keyword.
    - We should use keyword argument rather than Hash as a writing method.
    - It is unlikely to use the second argument.

### Aws::S3::Types

- It is using type alias a lot.
   - It conveys the context of the type, such as `Types::bucket`, better than `::String`.
   - Keep the code simple by not implementing type name resolution and it entrust to the rbs side.
- Structures is actually a `Struct`, but use class and attr_accessor to keep it simple.
- Enum is represented by union type of string literal


pocke:
Thanks for your PR. The generator script looks cool! :heart_eyes_cat: 
I'll review it. :rocket:

ksss:
The problem with the code generator created in this PR is too powerful.

This code generator can eventually provide rbs files for all aws-sdk gems, but if we provide rbs for every gem, the number of rbs will be 294, which is very large and will increase the management cost of this repository.

I have another idea.
The aws-sdk gems have some auto-generated files.

https://github.com/aws/aws-sdk-ruby/blob/9d29a70cd90482d34b4ecf1e0800a7f0ac1035f4/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/code_builder.rb#L53-L85

Conversely, it may be possible to parse the generated gem code to automatically generate rbs. For example, as you know [rbs_rails](https://github.com/pocke/rbs_rails) to auto-generate rbs files only for the aws-sdk gems that are already installed in user's application.

In this case, the management cost for gem_rbs_collection is low because I only need to manage the generator in a separate tool, but there is a concern of confusion for users because of the distribution of tools.
And, user can't manage the rbs-collection either.

What should we think about this issue?

pocke:
Thanks for sharing your concern! Sorry for the late reply.

I summarize the problem and the solution ideas to clear them.

## Problem

* Syncing RBS with many versions and many gems
  * We need to generate a new version of RBS for the corresponding version of the gem.
  * It is the same as other gems, but
    * `aws-sdk-*` gems are really actively released (for example, the current version of aws-sdk-s3 is 1.104.0). So we have to release RBS frequently.
    * There're 294 `aws-sdk-*` gems. It requires a huge cost to maintain.
* Managing the generator script
  * We also need to maintain the generator script.
  * I'm not sure how difficult managing it is.

## Solution Ideas

I have three solution ideas, including your proposal.

### Manage it in ruby/gem_rbs_collection

#### How

Merge this PR, and sync the generated RBSs with `aws-sdk-*` gems.

#### Pros

* Easy to start
* The users can be on the `rbs collection` way.

#### Cons

* The maintenance cost

### Make a third-party gem, like rbs_rails

#### How

As @ksss proposed, make a generator (e.g. `rbs_aws`), then the users install the tool and use it directly.

#### Pros

* No (or a few) maintenance cost for maintainers of gem_rbs_collection

#### Cons

* The users cannot be on the `rbs collection` way.
* The users need to generate RBSs of `aws-sdk-*` gems themselves.
  * It means the maintenance cost is passed on to the users from us.

### Manage it in aws/aws-sdk-ruby

#### How

Move the generator script to aws/aws-sdk-ruby.

 `aws-sdk-*` gems will include the generated RBSs under `sig/` directory of published gems.

#### Pros

* No maintenance cost for maintainers of gem_rbs_collection
* The users can be on the `rbs collection` way.

#### Cons

* AWS SDK Ruby team will have the maintenance cost.
* We need to negotiate with the AWS SDK Ruby team.

## Summary


Personally I think the first idea, managing it in this repository, is best for now. However we should move on to the last idea, managing it in aws/aws-sdk-ruby repo, in the future.

We can reduce the maintenance cost.
For example, not all versions and gems are required. Supporting major gems and the latest versions is enough.

In the future, I'd like to move the generator script to aws/aws-sdk-ruby. I think the operation experience of the genrator script will make negotiation and introdution easier.

So I'd like to try introducing the generator script on this repository. What do you think?

pocke:
I think the second idea is not appropriate for this case. The generated RBSs do not depend on user's code, so the RBS should be distributed from somewhere, but not generated by user.
But it will be a good solution for temporary use case. For example, generating RBS before releasing RBS in this repo.


——


By the way, I just think up another solution idea, making a collection repo only for aws-sdk-* gems. For example, ksss/aws_rbs_collection gh repo.
It is rbs collection compatible, flexibly changing, easy to automate. I didn’t consider it deeply, but it looks good solution.

ksss:
Thank you for awesome summarizing!

Finally, I also think that using `ruby/gem_rbs_collection` is a good way to start.
I think that all RBS generated by the generator should be managed only by the rake task.
There may be some delay in the release, but I don't think it will be a critical problem since users have the option to use the generator directly.

I think `ksss/aws_rbs_collection` idea seems to confuse the user.
Users have to wonder every time if they should PR the RBS of the aws-sdk-* gem to `ruby/gem_rbs_collection` or to `kss/aws_rbs_collection`.

pocke:
Thanks for your reply :+1: I'll review and merge this PR if there're no problems :rocket:


> I think ksss/aws_rbs_collection idea seems to confuse the user.

Right, it will confusion to the users.

If multiple collection sources are widely used in the future, we can reconsider this solution. But probably it is not appropriate for now.

pocke (commented):
The code looks good to me, thanks!

But I have a request to change, how about adding new entries to .gitignore for `Gemfile.lock` and the downloaded JSON file?

ksss:
@pocke Sorry some changes.
How about https://github.com/ruby/gem_rbs_collection/pull/76/commits/5007cf254d1c616058ccc2e9787bef75c4da0032 ?

In anticipation of the addition of other aws-sdk-* gems, I moved the Rakefile and generator to root.
Also, temporary files are now placed in the tmp directory.

ksss:
@pocke Sorry some changes.
How about https://github.com/ruby/gem_rbs_collection/pull/76/commits/b9022549a10990e72ba25661521cc6f3c9af6cba ?

In anticipation of the addition of other aws-sdk-* gems, I moved the Rakefile and generator to root.
Also, temporary files are now placed in the tmp directory.

pocke:
> In anticipation of the addition of other aws-sdk-* gems, I moved the Rakefile and generator to root.
Also, temporary files are now placed in the tmp directory.

Thanks, I think it is acceptable :+1:

PR #88, opened 2021-11-13
Nokogiri: Make block arguments optional

kymmt90:
Nokogiri::XML::Document has several methods to create nodes:

- [#create_element](https://github.com/sparklemotion/nokogiri/blob/v1.11.1/lib/nokogiri/xml/document.rb#L90-L121)
- [#create_text_node](https://github.com/sparklemotion/nokogiri/blob/v1.11.1/lib/nokogiri/xml/document.rb#L123-L126)
- [#create_cdata](https://github.com/sparklemotion/nokogiri/blob/v1.11.1/lib/nokogiri/xml/document.rb#L128-L131)
- [#create_comment](https://github.com/sparklemotion/nokogiri/blob/7be6f04aa2700e818f8a3bfe82801b5bd6e8c4f4/lib/nokogiri/xml/document.rb#L133-L136)

Those methods can be passed a block but it is not required, so those
requirements should be relaxed.

### DIFF
--- a/gems/nokogiri/1.11/_test/test.rb
+++ b/gems/nokogiri/1.11/_test/test.rb
@@ -31,3 +31,15 @@ end
 doc.search('nav ul.menu li a', '//article//h2').each do |link|
   puts link.content
 end
+
+# Create nodes
+doc.create_element('h1', 'hello') { |e| puts e }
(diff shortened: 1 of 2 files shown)


pocke (approved):
Thanks!

PR #89, opened 2021-11-18
Support Aws::S3::Client options

ksss:
## Problem

`Aws::S3::Client` is not able to support optional arguments.

```rb
Aws::S3::Client.new(
  region: 'ap-test-1' # => Unexpected keyword argument
)
```

## Solution

Any `Aws::*::Client` has many plugins.
Then, plugins defined options for `.new`.
Extract this option from `Aws::*::Client` class and define it as a keyword argument.
It is not exactly a keyword argument, but it practically works as a keyword argument.


ksss:
I guess I'll have to think about it some more...

.gitmodules
```diff
@@ -58,3 +58,6 @@
 [submodule "gems/zengin_code/1.0/_src"]
 	path = gems/zengin_code/1.0/_src
 	url = https://github.com/zengin-code/zengin-rb.git
+[submodule "gems/aws-sdk-core/3/_src"]
+	path = gems/aws-sdk-core/3/_src
+	url = https://github.com/aws/aws-sdk-ruby.git
```
ksss, line 63: Currently, `gems/aws-sdk-core` dir is just a place for _src.
Will it confuse the user?

ksss (commented):
Ready for review.

.gitmodules
pocke: >  Will it confuse the user?

Hmm, it will introduce a bit confusing behavior to `rbs collection`. By this change, `rbs collection` installs `aws-sdk-core` gem's RBS, but actually it is empty.
Could you reconsider it? 
Personally I think placing the submodule to aws-sdk-s3 directory is a good idea for now. WDYT?

ksss: As you said, `aws-sdk-core` doesn't seem to be a good idea.

I am planning to add other gems such as `aws-sdk-sqs` and `aws-sdk-kms` as well as `aws-sdk-s3` in the future.

The dependencies of the aws-sdk gem are as follows.
aws-sdk -> aws-sdk-resources -> aws-sdk-s3, ...

As an idea to reduce confusion, how about putting the source code submodule in the `aws-sdk` directory and make it dependent on `aws-sdk` to `aws-sdk-resources` and `aws-sdk-resources` to `aws-sdk-s3`? 

What do you think?

pocke: > I am planning to add other gems such as aws-sdk-sqs and aws-sdk-kms as well as aws-sdk-s3 in the future.

It's good news:+1: 

> As an idea to reduce confusion, how about putting the source code submodule in the aws-sdk directory and make it dependent on aws-sdk to aws-sdk-resources and aws-sdk-resources to aws-sdk-s3?

Did you mean putting submodule of aws-sdk-core to `gems/aws-sdk/0/_src` (`0` is tentative). I guess it has the same problem as I said previously until we write RBS for `aws-sdk` gem.


I have another idea, creating `generator/aws-sdk/` directory to the top level, then putting the submodule under the created directory. I think it will remove the confusion. We will probably need generator for other gen in the future, so the directory is meaningful I guess.

ksss: > Did you mean putting submodule of aws-sdk-core to gems/aws-sdk/0/_src (0 is tentative). 

Sorry for the confusion.
My plan is that put a submodule to https://github.com/aws/aws-sdk-ruby in `gems/aws-sdk/0/_src` and write a dependency to `aws-sdk-s3`, `aws-sdk-sqs` and so on in `gems/aws-sdk/0/manifest.yaml`.

> I have another idea, creating generator/aws-sdk/ directory to the top level

Sounds grate idea ✨ 
I'm going to do it this way!

ksss:
- Added `generators` directory
- Updated signature of aws-sdk-s3
- Wrote documentation of generators/aws-sdk

pocke (approved):
Looks good to me, thanks! And sorry for the delay to review:pray:

PR #90, opened 2021-11-21
Support Duration for {Date,Time}#{+,-}

ksss:
## Problem

The following code will fail during type checking (steep).

```rb
Date.today - 1.day
#=> (UnresolvedOverloading) Cannot find compatible overloading of method `-` of type `::Date`
```

## Why

Following methods are overwrote by active_support.

- `Date#+`
- `Date#-`
- `Time#+`
- `Time#-`

However, these methods have been omitted from the type definition for the following reasons and processing.

https://github.com/ruby/gem_rbs_collection/blob/88e86e0b67262f9ab6244a356e81dd9ca8c55b37/gems/railties/6.0/_scripts/generate.rb#L84-L85

## Solution

These methods should support `ActiveSupport::Duration` object for argument.
I suggest not using alias and using overload to solve the problem.

### DIFF
--- a/gems/activesupport/6.0/activesupport-generated.rbs
+++ b/gems/activesupport/6.0/activesupport-generated.rbs
@@ -2454,9 +2454,15 @@ class Date
 
   alias plus_without_duration +
 
+  def +: (ActiveSupport::Duration other) -> self
+       | ...
+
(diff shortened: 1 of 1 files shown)


pocke (approved):
LGTM :+1:

PR #91, opened 2021-11-23
Add methods generated by `matter_accessor`

ksss:
## Problem

`matter_accessor` https://github.com/rails/rails/blob/c7a6ec51bf097eb3f307001c1142a8565e7a1450/activesupport/lib/active_support/core_ext/module/attribute_accessors.rb#L207
generate accessor method.

However in activerecord-generated.rb, methods defined by `mattr_accessor` did not generate a type definition.

## Solution

I generated the type definition for the methods defined by `mattr_accessor` with the following code.

```rb
class Module
  prepend Module.new {
    def mattr_accessor(*syms, instance_reader: true, instance_writer: true, instance_accessor: true, **key)
      syms.each do |sym|
        puts "[#{self}] attr_accessor self.#{sym}: untyped"
        if instance_accessor
          if instance_reader
            puts "[#{self}] def #{sym}: () -> untyped"
          end
          if instance_writer
            puts "[#{self}] def #{sym}=: (untyped) -> untyped"
          end
        end
      end
      super
    end
  }
end
```

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -39,7 +39,6 @@ module ActiveRecord
     extend ::ActiveSupport::Callbacks::ClassMethods
 
     def self.abstract_class=: (bool) -> void
-    def self.default_timezone=: (untyped) -> untyped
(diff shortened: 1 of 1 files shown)


pocke (approved):
Thanks!

PR #92, opened 2021-11-27
Add methods generated by class_attributes for ActiveRecord::Base

ksss:
`Class#class_attributes` generate some methods on `ActiveRecord::Base` also.

I wrote method signature by following hook script.
Then, I extracted only the methods related to `ActiveRecord::Base`.

```rb
class Class
  prepend Module.new {
    def class_attribute(
      *attrs,
      instance_accessor: true,
      instance_reader: instance_accessor,
      instance_writer: instance_accessor,
      instance_predicate: true,
      **
    )
      attrs.each do |name|
        # skip internal attribute
        next if name.to_s.start_with?("_")
        next if name == :attribute_type_decorations
        next if name == :attributes_to_define_after_schema_loads
        puts "[#{self}] def self.#{name}: () -> untyped"
        puts "[#{self}] def self.#{name}?: () -> bool" if instance_predicate
        puts "[#{self}] def self.#{name}=: (untyped) -> untyped"
        if instance_reader
          puts "[#{self}] def #{name}: () -> untyped"
          puts "[#{self}] def #{name}?: () -> bool" if instance_predicate
        end
        if instance_writer
          puts "[#{self}] def #{name}=: (untyped) -> untyped"
        end
      end
      puts
      super
    end
  }
end
```

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -87,6 +87,142 @@ module ActiveRecord
     def errors: () -> untyped
     def []: (Symbol) -> untyped
     def []=: (Symbol, untyped) -> untyped
+
+    # class_attributes
+    def self.defined_enums: () -> untyped
+    def self.defined_enums?: () -> bool
+    def self.defined_enums=: (untyped) -> untyped
+    def defined_enums: () -> untyped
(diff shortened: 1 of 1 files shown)


pocke:
Thanks!

PR #95, opened 2021-12-12
Use rbs collection feature for testing

pocke:


pocke:
It has been achieved in #502

PR #106, opened 2021-12-28
Add aws-sdk-sqs gem

ksss:
I also added aws-sdk-core gem.
The aws-sdk-core currently has only a very small definition.
However, I believe it will be needed anyway.


gems/aws-sdk-s3/1/manifest.yaml
```diff
@@ -0,0 +1,2 @@
+dependencies:
+  - name: aws-sdk-core
```
pocke, line 2: I think the `manifest.yaml`s are not necessary.

`aws-sdk-s3` and `-sqs` declare the dependency to aws-sdk-core in the gemspec. ref: https://rubygems.org/gems/aws-sdk-sqs
`manifest.yaml` is a file to declare the dependencies that don't appear in the gemspec, such as standard libraries, so `aws-sdk-core` is not necessary to put in this place in this case.

ksss: Thanks 🎍
I understand about manifest.yaml.
I'll fix it later.

pocke (approved):
LGTM :star2:

PR #110, opened 2022-01-05
Support nested input argument for aws-sdk-*

ksss:
Currently, there is an error in the case of using Hash with nested input arguments such as `Aws::S3::Client#complete_multipart_upload`.

https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#complete_multipart_upload-instance_method

As a feature of aws-sdk, `structure` had two implications.

- If the input argument is a `structure`, it expects a Hash
- If the output is a `structure`, return with a dedicated class instance

In the existing rbs type, the input argument was also type-checked as a dedicated class instance.

I had to make some major changes to the generator to fix this problem.
(*.rbs file was generated automatically.)

Also, I added `_scripts/test` file to support CI.


ksss:
I'm thinking about the following problem...

```rbs
# `cors_rules` is conflicting!

# input
def put_bucket_cors: (… cors_configuration: cors_configuration, …) -> Aws::EmptyStructure
type cors_configuration = { cors_rules: cors_rules }
type cors_rules = ::Array[cors_rule]
type cors_rule = { ... }

# output
class GetBucketCorsOutput
  attr_accessor cors_rules: cors_rules
End
type cors_rules = ::Array[CORSRule]
class CORSRule
  ...
End
```

ksss:
Now on supported all input arguments!

ksss:
Dear Reviewer.

I'm planning to make even bigger changes to the generator based on this change.

So, I'm concerned about the extra burden on reviewers.

If reviewing the generator is too much work, for example, I can create a separate repository for the generator only, such as `ksss/aws-sdk-rbs-generator`, and send a pull request only RBS file to this repository.

What do you think?

pocke:
Thanks a lot for your contribution! 

> So, I'm concerned about the extra burden on reviewers.
>
> If reviewing the generator is too much work, for example, I can create a separate repository for the generator only, such as ksss/aws-sdk-rbs-generator, and send a pull request only RBS file to this repository.

I have the same concern. Actually, I'm spending much time reviewing PRs on this repository and 
it causes delays in review. I think I need to reduce the cost to avoid burning out.
So I appreciate your suggestion, but I'm not sure it is the best solution.

If we separate the repository, it will introduce confusion to new contributors perhaps. For example, a new contributor may modify the RBS in this repo directly.

So I suggest another approach, I'd like to promote you to a maintainer of `aws-sdk-*` RBSs. 
If you accept it, the following things will be changed.

* I'll merge your PR for maintained gems with a rough review.
* You also will be responsible for reviewing PRs to maintained gems.


What do you think? If you'd like to become a maintainer, let's start it from this PR!

By the way, I'm thinking I'll change just operations but will not configure anything, such as collaborator and code owner features for now.
I'd like to improve the operation in the future like DefinitelyTyped, so I'll investigate to find the solution.

pocke:
I completely understand CODEOWNERS feature :smile:, so I'm working to enable it.

ksss:
@pocke 

Thanks for the great suggestion.

I agree to be the maintainer of `gems/aws-sdk-*` and `generators/aws-sdk` in gem_rbs_collection.

I have already created https://github.com/ksss/aws-sdk-rbs-generator.
But, I would like to have this code merged into `gem_rbs_collection/generators/aws-sdk` and discard https://github.com/ksss/aws-sdk-rbs-generator.

pocke:
Thanks for your reply!


I'll merge this PR :+1:
I'll set up the CODEOWNERS feature and invite you as an outside collaborator this week. Please wait a while :pray:

PR #111, opened 2022-01-17
Fix `ActiveModel#validates_with`; block argument can be omitted

takahashim:
Without this fix, I got `[error] The method cannot be called without a block` with Validator class.


pocke:
Well, I encountered a bug in GitHub... I've clicked the merge button then the commit has been merged into the main branch, but it didn't change the PR status to "merged".
So I'll close this PR unfortunately, sorry for that and really thanks for your pull request!

PR #113, opened 2022-01-23
Add signatures for aws-sdk-core

ksss:
`Aws::Errors::ServiceError` is the base class for all errors returned by an Amazon Web Service.
`Aws::Resources::Collection` is the base class for all resource collection class.


pocke (approved):
Thanks!

PR #115, opened 2022-01-24
Add an RBS file for bcrypt-ruby

koukikitamura:
## Description
This PR adds an RBS file for [bcrypt-ruby](https://rubygems.org/gems/bcrypt-ruby).

## Checklist

- [x] Generate files with the `bin/init_new_gem` command.
- [x] Write standard libraries used in the gem to a manifest.yaml
- [x] Pass type checks with Steep through general usage of the gem.


pocke (approved):
Thanks!

PR #118, opened 2022-02-06
Reimplement generator for aws-sdk

ksss:
# What

## Merge [ksss/aws-sdk-rbs-generator](https://github.com/ksss/aws-sdk-rbs-generator)

As discussed in https://github.com/ruby/gem_rbs_collection/pull/110, I was preparing a big change about the generator for aws-sdk.
This is ksss/aws-sdk-rbs-generator.
This is a generator that completely replaces the `generators/aws-sdk`.

This PR add new generator in `generators/aws-sdk-rbs-generator`.
And delete `generators/aws-sdk`.

# Why

## Support `Client#wait_until`

Added support for `Client#wait_until`, which is a method that outputs differently depending on the argument.

## Support Resource classes

This change will allow us to generate signatures for the resource class as well as the traditional client class.
The resource class is, for example, the `Aws::S3::Resource` class.

```rb
resource = Aws::S3::Resource.new
obj = resource.bucket('a').object('b')
#=> We will be able to check that obj is an instance of Aws::S3::Object !
```

# How to operate

The generator will always overwrite and regenerate the target file.

Whenever an auto-generated file needs to be modified, the generator needs to be modified.

For classes that are not auto-generated, simply place the files under `gems/aws-sdk-*/`. This process is exactly the same as [aws-sdk-ruby](https://github.com/aws/aws-sdk-ruby).


pocke (approved):
Thanks :rocket:

PR #121, opened 2022-02-14
Fix stub_responses signature for aws-sdk

ksss:
The `stub_responses` option can also accept a hash, but it can also stub all operations and errors, which is expected to be very difficult to express accurately.

Therefore, it should be untyped once and not become an error.

See also: https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ClientStubs.html


ksss:
Fix EC2::Client also since https://github.com/ruby/gem_rbs_collection/pull/120

pocke (approved):
Thanks!

PR #122, opened 2022-02-18
Fix type error on resource class `#new`

ksss:
Fix the problem that the following valid code causes a type error.

```rb
Aws::S3::Bucket.new(name: 'a')
#=> More keyword arguments are required: client(Ruby::InsufficientKeywordArguments)
```

The client option is very long and duplicated.
Also, since this is a low-level API, the type is not that important.
If we want to make the type strict, we can use the `client` option,
otherwise, make the type untyped to avoid errors.


pocke (approved):
:+1:

PR #123, opened 2022-02-19
Add signature for customizations in s3

ksss:
There are some non-automatically generated files in S3, and I have added a signature for these files.
In particular, `Aws::S3::Object#presigned_url` and `Aws::S3::Object#presigned_url` are a major APIs that are also used in rails.

https://github.com/rails/rails/blob/15a2ff1d5bc9dcc3b120c31180c4c6a3d7d79ff3/activestorage/lib/active_storage/service/s3_service.rb#L117-L125


pocke (approved):
:rocket:

PR #125, opened 2022-02-20
Update steep

ksss:
steep v0.46.0 has a little bugs.
https://github.com/soutaro/steep/pull/444

We can expand the scope of what we can test by updating the steep.

I confirmed that it works by running `$ bin/test $(find gems -name '_scripts' | xargs dirname)`.


pocke (approved):
Thanks!

PR #132, opened 2022-03-19
Fix `unscoped` in ActiveRecord; allow without block

takahashim:
`unscoped` class method in ActiveRecord can be called without block.

ref. https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Default/ClassMethods.html#method-i-unscoped

So this PR fixes it.


takahashim:
I updated my fix to change `activerecord.rbs` file only, not `activerecord-generated.rbs` file.

pocke (approved):
Thanks! 👍

PR #134, opened 2022-03-20
Fix `TimeWithZone`; support `DateAndTime::Calculations`

takahashim:
`ActiveSupport::TimeWithZone` supports `DateAndTime::Calculations` through method_missing.

```ruby
> Time.zone.now.class
=> ActiveSupport::TimeWithZone
> Time.zone.now.yesterday
=> Sat, 19 Mar 2022 02:16:45.979101000 UTC +00:00
```

### DIFF
--- a/gems/activesupport/6.0/activesupport.rbs
+++ b/gems/activesupport/6.0/activesupport.rbs
@@ -74,6 +74,8 @@ module ActiveSupport
   end
 
   class TimeWithZone
+    include DateAndTime::Calculations
+
     # class_eval
     def year: () -> Integer
     def mon: () -> Integer


ksss (approved):
👍

pocke:
Thanks for your PR!

I understand the behavior as the following:

* AS::TimeWithZone delegates to Time via method_missing
  * https://github.com/rails/rails/blob/52f1f31c410f535032aa55401e0da583a9e768c0/activesupport/lib/active_support/time_with_zone.rb#L571
* Time includes DateAndTime::Calculations
  * https://github.com/rails/rails/blob/52f1f31c410f535032aa55401e0da583a9e768c0/activesupport/lib/active_support/core_ext/time/calculations.rb#L11-L12
* RBS doesn't support delegation

So we need to include `DateAndTime::Calculations` to `AS::TimeWithZone`. 


I think we need to reconsider this implementation but this workaround is ok for now.

pocke (approved):
🚀

PR #135, opened 2022-03-20
Fix ActiveRecord; allow omission of some blocks

takahashim:
In CollectionProxy of ActiveRecord, we can use `build`, `create` and `create!` without blocks.

```ruby
person.pets.build(name: 'Fancy-Fancy')
# => #<Pet id: nil, name: "Fancy-Fancy", person_id: 1>
```

https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-build

This PR overrides the current definition so that there is no need for the block.


pocke (approved):
👍

PR #136, opened 2022-03-20
Fix ActiveRecord; allow omission of blocks in default_scope

takahashim:
This PR is similar to #135 and allows class method blocks to be omitted.

In this case, `default_scope` is changed.


pocke (approved):
Thanks!

PR #137, opened 2022-03-20
Add some module methods for `Aws` in aws-sdk-core

takahashim:
This PR add some methods for Aws root module: `config` and `use_bundled_cert!`

https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws.html

(I found some other methods, but I'm not sure how to use them...)

### DIFF
--- a/gems/aws-sdk-core/3/aws.rbs
+++ b/gems/aws-sdk-core/3/aws.rbs
@@ -1,0 +1,4 @@
+module Aws
+  def self.config: () -> Hash[Symbol, untyped]
+  def self.use_bundled_cert!: () -> String
+end


gems/aws-sdk-core/3/aws.rbs
```diff
@@ -0,0 +1,4 @@
+module Aws
+  def self.config: () -> Hash[Symbol, untyped]
```
ksss, line 2: More precisely, it seems to be `attr_reader`.
https://github.com/aws/aws-sdk-ruby/blob/62e0113f7b9efe66bb77318ae85b147ed462a55c/gems/aws-sdk-core/lib/aws-sdk-core.rb#L119
```suggestion
  attr_reader self.config: Hash[Symbol, untyped]
```

ksss (commented):
LGTM.
But, Please name the file `aws-sdk-core.rbs` instead of `aws.rbs`.

takahashim:
Thanks for your comment!　I'll fix the points you mentioned.

takahashim:
Fixed!

ksss (approved):
Thank you ✨

ksss:
@pocke Can I merge them from now on?

PR #138, opened 2022-03-20
Fix ActiveRecord; allow omission of blocks in some callbacks

takahashim:
This PR is also similar to #135 and allows class method blocks to be omitted.

These methods are callbacks for transactions.

https://api.rubyonrails.org/v7.0.2.3/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_create_commit


pocke (approved):
👏

PR #139, opened 2022-03-20
Add paranoia gem

takahashim:
This PR adds rbs files of [paranoia](https://github.com/rubysherpas/paranoia) gem.


pocke (approved):
Thanks! 🪂

PR #140, opened 2022-03-20
Add a test for ActiveJob

takahashim:
There was no test for ActiveJob, so I wrote one.

Also, there seemed to be class methods that were not defined in `activejob-generated.rbs`, so I added `activejob.rbs` file.


pocke (approved):
🛴

PR #141, opened 2022-03-22
Add sentry-ruby gem

takahashim:
This PR adds rbs files of [sentry-ruby](https://github.com/getsentry/sentry-ruby/tree/master/sentry-ruby) gem.


pocke (approved):
Thanks!

PR #159, opened 2022-05-22
Add link to ruby/rbs on README

ybiquitous:
I think it's helpful to add a link to the [`ruby/rbs`](https://github.com/ruby/rbs) repository, especially for people unfamiliar with RBS.


pocke (approved):
Thanks!

PR #164, opened 2022-06-03
Add RBS for ActiveDecorator gem

fugakkbn:
This PR adds rbs files of [ActiveDecorator](https://github.com/amatsuda/active_decorator) gem.

<hr>

I opened a PR for the first time.
I would like to contribute at every opportunity in the future.

However, since I am new to this field, I am sure there will be many oddities and mistakes.
I would be grateful for any suggestions or advice you can give me.
Thanks for your maintenance.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -79,3 +79,6 @@
 [submodule "gems/rolify/6.0/_src"]
 	path = gems/rolify/6.0/_src
 	url = https://github.com/RolifyCommunity/rolify.git
+[submodule "gems/active_decorator/1.4/_src"]
+	path = gems/active_decorator/1.4/_src
+	url = https://github.com/amatsuda/active_decorator.git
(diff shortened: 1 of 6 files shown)


gems/active_decorator/1.4/manifest.yaml
```diff
@@ -0,0 +1,7 @@
+# manifest.yaml describes dependencies which do not appear in the gemspec.
+# If this gem includes such dependencies, comment-out the following lines and
+# declare the dependencies.
+# If all dependencies appear in the gemspec, you can remove this file.
```
pocke, line 4: Could you remove this file? It looks not necessary.
`manifest.yaml` is needed to list standard libs dependencies, and the `active_decorator.rbs` doesn't have the dependencies. So it is not necessary.

fugakkbn: Thanks for your review!
I removed `manifest.yaml’ file.
I missed comments in codes😅

pocke (approved):
Thanks!

PR #165, opened 2022-06-05
Allow optional block for `#find_or_create_by`

fugakkbn:
`ActiveRecord::Relation#find_or_create_by(!)` type definition requires a block.
If we try to run without blocks, we will get the following error:

```
[error] The method cannot be called without a block
│
└     users.find_or_create_by!(nickname: fugakkbn)
```

However, since there are cases where it is called without a block, this PR makes the change to make the block optional.


pocke (approved):
Thanks!

PR #168, opened 2022-06-12
Allow some methods of `ActiveRecord::Relation` to accept optional block arguments

fugakkbn:
The following cases will result in an error when `steep check` is run:
```ruby
class User < ApplicationRecord
  has_many :messages
  
  def replied?
    messages.any? {|message| message.replied_at.exists?}
  end
end

class Message < ApplicationRecord
  belongs_to :user
end
```

```
$ bundle exec steep check

[error] The method cannot be called with a block
│ Diagnostic ID: Ruby::UnexpectedBlockGiven
│
└     messages.any? {|message| message.replied_at.exists?}
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The cause is that `ActiveRecord::Relation#any?` is a type definition that does not accept blocks.
So, this PR modifies the type definition so that `ActiveRecord::Relation#any?` can optionally accept blocks.
In addition, the return types of methods and blocks are also modified.

Similarly, modify `#none?`, `#one?`, and `#many?`.

### DIFF
--- a/gems/activerecord/6.0/activerecord-generated.rbs
+++ b/gems/activerecord/6.0/activerecord-generated.rbs
@@ -19731,16 +19731,16 @@ module ActiveRecord
     def empty?: () -> untyped
 
     # Returns true if there are no records.
-    def none?: () -> untyped
+    def none?: () ?{ (untyped) -> bool } -> bool
 
     # Returns true if there are any records.
-    def any?: () -> untyped
+    def any?: () ?{ (untyped) -> bool } -> bool
 
     # Returns true if there is exactly one record.
-    def one?: () -> untyped
+    def one?: () ?{ (untyped) -> bool } -> bool
 
     # Returns true if there is more than one record.
-    def many?: () -> untyped
+    def many?: () ?{ (untyped) -> bool } -> bool
 
     # Returns a stable cache key that can be used to identify this query.
     # The cache key is built with a fingerprint of the SQL query.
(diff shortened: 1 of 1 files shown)


gems/activerecord/6.0/activerecord-generated.rbs
```diff
@@ -19731,16 +19731,16 @@ module ActiveRecord
     def empty?: () -> untyped
 
     # Returns true if there are no records.
-    def none?: () -> untyped
+    def none?: () ?{ (untyped) -> bool } -> bool
```
pocke, line 19734:
```suggestion
    def none?: () ?{ (untyped) -> boolish } -> bool
```

Could you change the returned value of the block to `boolish`? `boolish` is an alias of `top`, and `bool` is an alias of `true | false`.
In this case, the block can return any values other than `bool`, such as `nil`. So it should be `boolish` instead of `bool`.

We need to update the other methods in the same way also.

ref: https://github.com/ruby/rbs/blob/b16ea15a11cf9cdc098da94080268e4ec0767955/docs/syntax.md#bool-or-boolish

pocke (changes requested):
Thanks for your pull request!
I left a comment about `bool` usage. Could you check the comment?

gems/activerecord/6.0/activerecord-generated.rbs
fugakkbn: Thanks for your review!

I changed the return value of the block to `boolish`.
Thank you for your kind explanation.

pocke (approved):
👍

PR #172, opened 2022-06-19
Add #count type definitions to activerecord

fugakkbn:
Added type definition for `ActiveRecord::Calculations#count`.
Also, made it possible to accept blocks.


pocke (approved):
👍

PR #176, opened 2022-06-27
Add rails-generator for rails gems

ksss:
## Proposal

I have implemented a generator for rails gems.
I propose to use `rails-generator` to manage RBS for rails gems.
Simply run the `rake` command to generate an RBS with the following gems and versions from rails gem source code.

- gems
  - actionpack
  - actionview
  - activejob
  - activemodel
  - activerecord
  - activestorage
  - activesupport
  - railties
- versions
  - 6.0
  - 6.1
  - 7.0

I have not included it in this PR because generating all RBSs would result in very large differences.
Please review the generator only in this PR.
The generated RBS can be viewed at https://github.com/ksss/gem_rbs_collection/tree/rails-generator-full .

## Problem

I know there is a `gems/railties/6.0/_scripts/generate.rb`.

However, currently `*-generated.rbs` file is edited without generate.rb and the definition cannot be overwritten.

Also, static analysis cannot handle dynamically added methods and mixins.

## Solution

So, based on `generate.rb`, I have further improved it, added various features, added tests, and made it into a gem package.

https://github.com/ksss/orthoses
https://github.com/ksss/orthoses-rails

These gems are like tapioca for sorbet.

This generator makes RBS as automated as possible and allows to add definitions by RBS file.

The following features are currently available for automation, and additional features are under development.

- Dynamic mixins, attr methods, and constants
- Inheritance via `DelegateClass()` like Tempfile
- Methods dynamically added by `class_attribute`, `mattr_accessor` and `config_accessor`.
- Delegate methods to `ActiveSupport::TimeWithZone` from `Time` by `method_missing`.
- Method definition by `delegate` when the original definition exists.
- Split railties mixin to `railties` for gem dependency.

We put these functions on the base of `$ rbs prototype rb` and merge RBS in a separate file.

## Roadmap

First We will add the RBS created by the generator in rails 7.0 to gem_rbs_collection.
Then replace the existing 6.0, 6.1 with the RBS created by the generator.
Once the functionality is added to the generator, We will regenerate it all together.
If more rails versions are added, we will generate them each time.

## Operation

If we want to add a new RBS definition, we consider an operation that makes changes to `rails-generator/known_sig` and regenerates it.
The reason why the RBS is necessary for the generator is that it needs to know the type of the target when delegating by `delegate` or `method_missing`.
We would like to put the details in `rails-generator/README.md`.

## Review point

- `generators/rails-generator/`
    - README.md
        - Operation document.
    - Rakefile
        - Main generator script using orthoses.
    - known_sig/*
        - Specified RBS.
    - tasks/*
        - Rake task for each gems.
    - config/*
        - Minimum files required to start a rails application.


ksss:
@pocke This is a big change, but I hope you can see it.

pocke:
Thanks for your PR and sorry for my absence. I'll review this PR this week.

generators/rails-generator/rbs_collection.yaml
```diff
@@ -0,0 +1,94 @@
...
+sources:
+  - name: ruby/gem_rbs_collection
+    remote: https://github.com/ruby/gem_rbs_collection.git
+    revision: main
+    repo_dir: gems
+
+# A directory to install the downloaded RBSs
+path: .gem_rbs_collection
+
+gems:
```
pocke, line 11: I think it is not the intended usage of `rbs collection`.

`rbs collection` is designed to manage RBS dependencies. It supports downloading and loading RBSs to use them by Steep and so on.

I think this `rbs_collection.yaml` should be used to develop this generator. Actually, I misunderstood this file is used for it.


This usage introduces impedance mismatch, such as the many ignored gems.

So, I'd like to avoid using `rbs_collection.yaml` for this purpose.
How about the following alternative approach?


* List library names by hands without `rbs collection`.
* Rename this file like `rbs_collection-all-stdlibs.yaml`
* Create a directory and put `Gemfile` and `rbs_collection.yaml` to load all libraries.
  * e.g. `mkdir all-stdlibs` and put a Gemfile that contains nokogiri, rack, and so on.

generators/rails-generator/tasks/action_pack.rake
```diff
@@ -0,0 +1,144 @@
...
+        sh "cp -a out/#{version}/action_dispatch #{export}"
+        sh "cp -a out/#{version}/abstract_controller.rbs #{export}"
+        sh "cp -a out/#{version}/abstract_controller #{export}"
+        sh "cp -a out/#{version}/mime.rbs #{export}"
+        sh "cp -a out/#{version}/mime #{export}"
+
+        sh "rm #{export}/action_controller/railtie.rbs"
+        sh "rm #{export}/action_dispatch/railtie.rbs"
+
+        Pathname(export).join("EXTERNAL_TODO.rbs").write(<<~RBS)
```
pocke, line 29: What is `EXTERNAL_TODO.rbs`? I couldn't understand what this file contains.
Could you describe the purpose of this file and clarify it by renaming this file, adding comments, or something?

pocke, line 29: Why the files are written in the `.rake` file? I think it is better if this file is placed as a `.rbs` file. It improves editor integrations and so on.

generators/rails-generator/README.md
```diff
@@ -0,0 +1,84 @@
...
+Rails gems has dependencies, so it loads all library code and outputs an RBS file.
+
+### Export task
+
+```
+$ bundle exec rake 7.0:active_support:export
+```
+
+From the `out` directory, select the necessary files for each gem and copy them to `export`.
+Generate a test script at the same time so that the entire directory can be replaced.
```
pocke, line 68: Why the `export` directory is necessary? I think it's enough that this task copies the files to `gems/GEM_NAME/` directly and runs the test in the directory.

pocke (commented):
I'm positive to merge this PR, but I left some comments. Could you confirm them?

---


How about adding a test to the generator is applied? It is useful that the test detects missing executions. Do you have any plan to add such a test?

.github/CODEOWNERS
```diff
@@ -10,3 +10,4 @@
 
 # Misc
 /generators/aws-sdk-rbs-generator @ksss
+/generators/rails-generator @ksss
```
pocke, line 13: How about assigning you (and @pocke) to `/gems/actionpack` and so on too? A generator change affects under the `gems` directory, so I think you should have permission to them also.

generators/rails-generator/Rakefile
```diff
@@ -0,0 +1,337 @@
...
+  Pathname(export).join('_scripts').tap(&:mkdir).join('test').write(<<~SHELL)
+    #!/usr/bin/env bash
+
+    # !!! GENERATED CODE !!!
+    # Please see generators/rails-generator
+
+    # set -eou => Exit command with non-zero status code, Output logs of every command executed, Treat unset variables as an error when substituting.
+    set -eou pipefail
+    # Internal Field Separator - Linux shell variable
+    IFS=$'\n\t'
```
pocke, line 227: Could you apply the same changes as https://github.com/ruby/gem_rbs_collection/pull/224?
It's my mistake. Sorry for the inconvenience.

generators/rails-generator/Gemfile.lock
```diff
@@ -0,0 +1,175 @@
+PATH
+  remote: _src
+  specs:
+    actioncable (7.0.3.1)
```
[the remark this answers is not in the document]
ksss: I wanted to read source code for `rbs prototype rb`, so I didn't separate Gemfile and used `git switch` to change version.

I can read the source code of the installed gem, so maybe I should stop using `_src` and separate the Gemfile. I'll give it a try.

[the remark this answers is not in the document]
pocke: > I wanted to read source code for rbs prototype rb, so I didn't separate Gemfile and used git switch to change version.



Well, I guess `rbs prototype rb` can read the Rails source code from `_src` even if the Gemfile is separated.

But I think reading the source code from the installed gems.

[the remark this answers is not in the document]
ksss: I have separated the Gemfile into `/gemfiles` by rails version.

generators/rails-generator/Gemfile
```diff
@@ -0,0 +1,15 @@
+source "https://rubygems.org"
+
+gem "rails", path: "_src"
```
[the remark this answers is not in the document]
ksss: Fixed to read from gem source code.

generators/rails-generator/rbs_collection.yaml
ksss: It is mainly used by orthoses to check for overlap with existing definitions.
Renamed to orthoses.yaml.

generators/rails-generator/tasks/action_pack.rake
ksss: Renamed to patch.rbs and added explanatory documentation to README.

generators/rails-generator/README.md
ksss: Removed `export` and `validate`.
Also, changed the install task to copy directly to `/gems`.
Also, added a test task that executes `_scripts/test`.

generators/rails-generator/Rakefile
ksss: Supported 👍

ksss:
@pocke 
Many thanks for the confirmation.
I have corrected the points raised and updated the code to the latest state. Please check it.

ksss:
@pocke 
What is the status of this?
I believe that PRs like following will be more manageable in the future by incorporating this change.

- https://github.com/ruby/gem_rbs_collection/pull/277
- https://github.com/ruby/gem_rbs_collection/pull/280
- https://github.com/ruby/gem_rbs_collection/pull/281

pocke:
I'll close this PR due to the lack of activities.

Issue #177, opened 2022-06-29
Non-overloading method of `#to_time` in `Time::` on ActiveSupport


clinejj:
After updating to the latest RBS/Steep versions (2.6.0 and 1.0.1 respectively), we are getting the following error after running `steep check`: 

```sh
.gem_rbs_collection/activesupport/6.0/activesupport-generated.rbs:6115:2: [error] Non-overloading method definition of `to_time` in `::Time` cannot be duplicated
│ Diagnostic ID: RBS::DuplicatedMethodDefinition
│
└   def to_time: () -> untyped
    ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

see:
https://github.com/ruby/gem_rbs_collection/blob/cd6db41310b911e1df3476246257510673f01b8a/gems/activesupport/6.0/activesupport-generated.rbs#L6110-L6116

This was added in RBS 2.6.0 at https://github.com/ruby/rbs/pull/1033 by @ksss, but the type definitions here have not been updated for ActiveSupport

ksss:
Good catch!
I'll fix soon!

PR #178, opened 2022-06-30
Fix duplicate definition

ksss:
Fix https://github.com/ruby/gem_rbs_collection/issues/177

In rails add `Time#to_time` from https://github.com/rails/rails/pull/28147 for compatibility.
gems/railties/6.0/_scripts/generate.rb output this signature.
But, In rbs 2.6.0 added `Time#to_time` signature in date library.
So, Remove the gem_rbs_collection side, since the expected signatures for both are the same.


pocke (approved):
Thanks!

PR #183, opened 2022-07-21
Add some type declaration to Redis gem

fugakkbn:
The current type of the first argument `key` of `Redis#set, get` is only `String`, but Symbol is also used as `key` in some cases.

```ruby
redis = Redis.new

redis.set(:mykey, 'hello world')  # => 'OK'
redis.get(:mykey)  # => 'hello world'
```

Therefore, modify each method so that Symbol can also be used as the first argument `key`.

<hr>

In some cases, the above methods are also used with `Redis.current`.
`Redis.current` returns a Redis instance.

```ruby
Redis.current.set(:mykey, 'hello world')  # => 'OK'
Redis.current.get(:mykey)  # => 'hello world'
```

The `Redis.current` type declaration did not seem to be there yet, so I will add it.

### DIFF
--- a/gems/redis/4.2/_test/get_set.rb
+++ b/gems/redis/4.2/_test/get_set.rb
@@ -5,3 +5,7 @@ redis.set("mykey", "hello world")
 
 redis.get("mykey")
 # => "hello world"
+
+Redis.current.set(:mykey, 'hello world')
+
+Redis.current.get(:mykey)
--- a/gems/redis/4.2/_test/get_set.rb
+++ b/gems/redis/4.2/_test/get_set.rb
@@ -6,6 +6,8 @@ redis.set("mykey", "hello world")
 redis.get("mykey")
 # => "hello world"
 
-Redis.current.set(:mykey, 'hello world')
+redis.set(:other_key, 'hello world')
+# => "OK"
 
-Redis.current.get(:mykey)
+redis.get(:other_key)
+# => "hello world"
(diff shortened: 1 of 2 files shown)


gems/redis/4.2/_test/get_set.rb
```diff
@@ -5,3 +5,7 @@
 
 redis.get("mykey")
 # => "hello world"
+
+Redis.current.set(:mykey, 'hello world')
```
pocke, line 9: https://github.com/redis/redis-rb/blob/d6626e6ef0c49374a888074652c4c9b2237dc8eb/lib/redis.rb#L39-L46

`Redis.current` is deprecated. So could you use `Redis.new` instead in this test code?


And the added type for `Redis.current` can be removed also. Personally, I think we do not need to add a type definition for a deprecated method actively. Because a deprecated method should not be used, so we can omit to add it to avoid misusing the deprecated method.

---


By the way, RBS does not have any rule for deprecated methods yet. I'd like to introduce something to mark a type deprecated, such as an annotation. But there is no implementation for now.

fugakkbn: Thank you for your review.

There are no rules regarding deprecated methods in RBS, and you basically have a policy of not adding types.
I understand.

I'll fix the test and remove the `Redis.current` type definition.

pocke (approved):
👍

PR #184, opened 2022-07-22
Add resource waiters class's RBS for aws-sdk gems

ksss:
I added resource waiters class's RBS for aws-sdk gems (e.g. `Aws::S3::Waiters::BucketExists`).
And update it's source also.


pocke (approved):
🚀

Issue #190, opened 2022-08-12
Process for generating types for new aws-sdk gems


G-Rath:
I'm using `aws-sdk-ssm` and would love to have types for that - I've seen there's a [generator](https://github.com/ruby/gem_rbs_collection/tree/main/generators/aws-sdk-rbs-generator) but I don't see how it correlates with the handful of `aws-sdk` gems that are in this repo? i.e. I don't see anywhere a list that has just the clients that are currently here, but then this repo doesn't already have every aws client.

What's the process for generating the types for new clients?

ksss:
@G-Rath 
Thank you for the feedback!
This is the first feedback and I am really pleased.
I created the PR. Please feel free to give me feedback on its use also.

https://github.com/ruby/gem_rbs_collection/pull/191

G-Rath:
@ksss thanks for the fast response and PR! it looks good to me - I've made a comment about one aspect that I think the docs should cover, but the types themselves look great!

ksss:
@G-Rath 
I have generated an RBS for `aws-sdk-ssm` https://github.com/ruby/gem_rbs_collection/pull/191 .
Feel free to let me know if you have any problems.

PR #191, opened 2022-08-13
Add aws-sdk-ssm

ksss:
Support https://github.com/ruby/gem_rbs_collection/issues/190


generators/aws-sdk-rbs-generator/Rakefile
```diff
@@ -21,6 +21,14 @@ task :write do |task, args|
...
-task :default do
-  Rake::Task['write'].invoke("S3", "SQS", "KMS", "EC2")
+task :update_src do
+  cd "_src" do
+    sh "git pull origin version-3"
+  end
 end
+
+task :generate_all do
+  Rake::Task['write'].invoke("S3", "SQS", "KMS", "EC2", "SSM")
```
G-Rath, line 31: Ah this looks like what I was expecting but failed to find - I think this should be included in the documentation you've added (i.e. that only services in this "array" are covered, and that it's here that new services should be added)

ksss: @G-Rath Thank you for reviewing.
How about with the additions to the [generators/aws-sdk-rbs-generator/README.md](https://github.com/ruby/gem_rbs_collection/pull/191/files#diff-7820d191b29cfc265b704758b3dca4fb22e907a62dd9775fa033cc81ebe2de68)?

G-Rath: I was thinking something more explicit i.e. "Add the client service to the list in the `generate_all` task".

(tbh its probably worth pulling that out into an array constant since it could grow quite large over time, and then you could say something like "Add the missing service to the `AWS_SERVICES` array and run `bundle exec rake`")

ksss: @G-Rath Looks Good, I've added that sentence. Thanks!
This generator is still in a phase of gradual expansion as we listen to feedback, but eventually we intend to eliminate the need to add to the list.

pocke (approved):

PR #193, opened 2022-08-21
Add RBS for enumerize gem

fugakkbn:
Add type definitions for [enumerize gem](https://github.com/brainspec/enumerize).
The type definitions to be added are those generated by the `rbs prototype rb` command.

```shell
$ rbs prototype rb gems/enumerize/2.5/_src/lib/**/*.rb >> gems/enumerize/2.5/enumerize-generated.rbs
```

Sorry for the small amount of tests...
I can't find any cases where other methods are used for projects, and this is the only one.

### DIFF
--- a/gems/enumerize/2.5/README.md
+++ b/gems/enumerize/2.5/README.md
@@ -1,4 +1,4 @@
-# RBS for config gem
+# RBS for enumerize gem
 
 The directory that defines the RBS files for the [enumerize](https://github.com/brainspec/enumerize) gem.
 
--- a/gems/enumerize/2.5/README.md
+++ b/gems/enumerize/2.5/README.md
@@ -1,0 +1,42 @@
+# RBS for config gem
+
+The directory that defines the RBS files for the [enumerize](https://github.com/brainspec/enumerize) gem.
+
+## Usage
+
+### enumerize method
+
+The most common use cases for this gem would be:
+
+```rb
+# user.rb
+require 'enumerize'
+
+class User
+  extend Enumerize
+
+  enumerize :role, in: %i(user admin)
+end
+```
+
+In this case, the RBS file should be written as follows:
+
+```
+# user.rbs
+class User
+  extend Enumerize
+  extend Enumerize::Base::ClassMethods
+end
+```
+
+Note that `Enumerize::Base::ClassMethods` is also necessary.
+
+## Why do we need additional `extend` to RBS file?
+
+Take the `enumerize` method as an example. Because this method is dynamically defined in the calling class.
+[enumerize/base\.rb at d4438ecf8f90a24b8fb0e84b9d4338ae4a236352 · brainspec/enumerize](https://github.com/brainspec/enumerize/blob/d4438ecf8f90a24b8fb0e84b9d4338ae4a236352/lib/enumerize/base.rb#L6)
+
+Therefore, simply writing `extend Enumerize` in the RBS file does not define the `enumerize` method.
+
+Currently, RBS doesn't have any features to include/extend modules dynamically, so this problem can occur with other methods.
+In such cases, check the source of the method definition and include/extend it in the RBS file.
--- a/gems/enumerize/2.5/_test/test.rb
+++ b/gems/enumerize/2.5/_test/test.rb
@@ -1,0 +1,7 @@
+require 'enumerize'
+
+class User
+  extend Enumerize
+
+  enumerize :role, in: %i(user admin)
+end
--- a/gems/enumerize/2.5/_test/test.rbs
+++ b/gems/enumerize/2.5/_test/test.rbs
@@ -1,0 +1,4 @@
+class User
+  extend Enumerize
+  extend Enumerize::Base::ClassMethods
+end
--- a/.gitmodules
+++ b/.gitmodules
@@ -82,3 +82,6 @@
 [submodule "gems/active_decorator/1.4/_src"]
 	path = gems/active_decorator/1.4/_src
 	url = https://github.com/amatsuda/active_decorator.git
+[submodule "gems/enumerize/2.5/_src"]
+	path = gems/enumerize/2.5/_src
+	url = https://github.com/brainspec/enumerize.git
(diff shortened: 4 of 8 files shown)


gems/enumerize/2.5/_test/test.rb
```diff
@@ -0,0 +1,7 @@
+require 'enumerize'
+
+class User
+  extend Enumerize
+
+  enumerize :gender, in: %i(man woman)
```
pocke, line 6: I'd like to avoid using "gender" as an example of an enum. Actually, gender is not binary. It is not appropriate for this example. Could you replace it with another example, such as `role` like [the README](https://github.com/brainspec/enumerize#usage)?

gems/enumerize/2.5/_test/test.rbs
```diff
@@ -0,0 +1,3 @@
+class User
+  extend Enumerize::Predicates
```
pocke, line 2: There is a diference between `test.rb` and `test.rbs`. In `test.rb`, `User` calls `extend` with `Enumerize`. But in `test.rbs`, `User` calls `extend` with `Enumerize::Predicates`.

I think this difference is confusing. When a model has `extend Enumerize`, the user should need to write only `extend Enumerize` to an rbs file. But the user needs to write `extend Enumerize::Predicates` actually with this patch.

pocke:
> I can't find any cases where other methods are used for projects, and this is the only one.

You can find examples from the README if you need them.
https://github.com/brainspec/enumerize

I guess the added example represents the main use case of this gem, so it is not necessary to add more test cases for the first step.

gems/enumerize/2.5/_test/test.rb
fugakkbn: Thank you for pointing this out. As you say, this example is not appropriate.
Fixed to use "Role" as an example.

gems/enumerize/2.5/_test/test.rbs
fugakkbn: Please let me consult with you.

When we do `extend Enumerize` on `test.rbs`, we get the following error

```
test.rb:6:2: [error] Type `singleton(::User)` does not have method `enumerize`
│
└   enumerize :role, in: %i(user admin)
    ~~~~~~~~~

Detected 1 problem from 1 file
```

When we execute `extend Enumerize`, `Enumerize::Base.enumerize` is dynamically added to the calling class. I think that, this is why the above error occurs.

In this case, in `test.rbs`, I think I have no choice but to write it like `extend Enumerize::Base::ClassMethods`, is there any other better way?

pocke: Thank you for the explanation.

Currently, RBS doesn't have any features to include/extend modules dynamically, so we cannot avoid differences between `.rbs` and `.rb`.

> In this case, in test.rbs, I think I have no choice but to write it like extend Enumerize::Base::ClassMethods, is there any other better way? As you say, it looks odd that test.rb and test.rbs are written differently.


I agree that we have no choice but to write the different `extend`. We cannot avoid this problem.

But I think we can reduce confusion among the end-users. Currently, the users are hard to know they need to write the different `extend`. The users will just write `extend Enumerize`, then they'll get the `Ruby::NoMethod` error. This is not a good experience.
So it needs something to help the users, such as documentation and/or a tool that automatically adds `extend`s.

How about adding a README file to `gems/enumerize/2.5/` to describe this problem, and writing a documentation comment to `Enumerize` module? It will be helpful to the users.

----


By the way, I'm not sure the current inheritance is the best way. How about adding `extend Enumerize` to the RBS file?

Currently, the rbs file only has `extend Enumerize::Base::ClassMethods`, but actually, the `User` class's inheritance is the following:

```ruby
require 'enumerize'

class User
  extend Enumerize
end

pp User.ancestors
# => [User, Enumerize::Base, Object, Kernel, BasicObject]
pp User.singleton_class.ancestors
# => [Enumerize::Base::ClassMethods::Hook,
#     #<Class:User>,
#     Enumerize::ModuleAttributes,
#     Enumerize::Predicates,
#     Enumerize::Base::ClassMethods,
#     Enumerize,
#     #<Class:Object>,
#     #<Class:BasicObject>,
#     Class,
#     Module,
#     Object,
#     PP::ObjectMixin,
#     Kernel,
#     BasicObject]
```

So technically the RBS file should be the following:

```rbs
class User
  include Enumerize::Base
  extend Enumerize
  extend Enumerize::ModuleAttributes
  extend Enumerize::Predicates
  extend Enumerize::Base::ClassMethods
  extend Enumerize::Base::ClassMethods::Hook
end
```


(Technically, it also should have `Enumerize::ActiveRecordSupport` and so on in some cases.)


I think the full `extend`s list is overkill. Because the main purpose of these modules is to define `enumerize` method, and the method is defined by `Enumerize::Base::ClassMethods`. So the current RBS file is enough for 99% of users I guess.

So most of the `extend`s are not necessary, but I think `extend Enumerize` is necessary. Because the Ruby file has `extend Enumerize` explicitly, so missing `extend Enumerize` in the RBS file looks odd. WDYT?

fugakkbn: Thank you for your very kind reply!

> How about adding a README file to gems/enumerize/2.5/ to describe this problem, and writing a documentation comment to Enumerize module? It will be helpful to the users.

Adding a README file would be a great way to do this. This way, as you say, it would reduce user confusion. Thanks for the suggestion!

---

> So most of the extends are not necessary, but I think extend Enumerize is necessary. Because the Ruby file has extend Enumerize explicitly, so missing extend Enumerize in the RBS file looks odd. WDYT?

I fully agree with you. I think `extend Enumerize` should be written in the RBS file as well. I think it is reasonable to adding `extend Enumerize` and the `Enumerize::Base::ClassMethods` that defines the `enumerize` method in the RBS file.

```
class User
  extend Enumerize
  extend Enumerize::Base::ClassMethods
end
```

I’ll fix `test.rbs` look like this and document its use in the README as well.
Thank you for your specific suggestions.

fugakkbn: Fixed `test.rbs` and added README.md.

gems/enumerize/2.5/README.md
```diff
@@ -0,0 +1,42 @@
+# RBS for config gem
```
pocke, line 1:
```suggestion
# RBS for enumerize gem
```

Is it a typo? 😄

pocke (approved):
Thanks!

PR #196, opened 2022-08-31
Inherit `Aws::EmptyStructure` if the member is empty.

ksss:
Mimics Ruby's inheritance structure.

```rb
Aws::KMS::Types::ConnectCustomKeyStoreResponse.superclass
#=> Aws::EmptyStructure
```

I modified under `generators/aws-sdk-rbs-generator` and ran `$ bundle exec rake` to generate RBS.


pocke (approved):
🚀

PR #199, opened 2022-09-09
Add RBS for holidays gem

nipe0324:
Add type definitions for https://github.com/holidays/holidays 8.4

The type definitions are generated by `rbs prototype rb` command as below.

```sh
$ rbs prototype rb gems/holidays/8.4/_src/lib/**/*.rb > gems/holidays/8.4/holidays-generated.rbs
```

Sorry for the few tests. I use two methods in my project.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -85,3 +85,6 @@
 [submodule "gems/enumerize/2.5/_src"]
 	path = gems/enumerize/2.5/_src
 	url = https://github.com/brainspec/enumerize.git
+[submodule "gems/holidays/8.4/_src"]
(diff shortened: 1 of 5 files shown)


pocke (approved):
Looks good to me, thanks!

By the way, probably we can simplify the generated rbs file because it has noisy literal types. Let's try fixing this problem if you'd like!

nipe0324:
> By the way, probably we can simplify the generated rbs file because it has noisy literal types. Let's try fixing this problem if you'd like!

I think so too. I'll try fixing it.

PR #202, opened 2022-09-11
Remove noisy auto generated rbs from holidays

nipe0324:
From https://github.com/ruby/gem_rbs_collection/pull/199

> By the way, probably we can simplify the generated rbs file because it has noisy literal types. Let's try fixing this problem if you'd like! 

I tried fixing it.
I may have deleted too much ...


pocke (approved):
Thanks! It looks good for the minimal type definitions.

PR #203, opened 2022-09-12
Add RBS for sorcery

m11o:
Add RBS file for sorcery v0.16
This PR added RBS file for often used methods because I don't understand it so much.
Please check it.

[Sorcery/sorcery](https://github.com/Sorcery/sorcery)

### DIFF
--- a/gems/sorcery/0.16/sorcery.rbs
+++ b/gems/sorcery/0.16/sorcery.rbs
@@ -15,7 +15,7 @@ module Sorcery
       def logged_in?: () -> bool
       def current_user: () -> (T | nil)
       def logout: () -> (nil | void)
-      def auto_login: (T, ?bool) -> T
+      def auto_login: (T, ?bool) -> void
       def redirect_back_or_to: (String url, ?Hash[(String | Symbol), String] flash_hash) -> (nil | void)
     end
   end
--- a/gems/sorcery/0.16/sorcery.rbs
+++ b/gems/sorcery/0.16/sorcery.rbs
@@ -10,7 +10,9 @@ module Sorcery
   module Controller
     module InstanceMethods[T]
       def require_login: () -> void
-      def login: [U] (*String credentials) ?{ (T, nil) -> U } -> (T | nil | U)
+
+      type failure_reason = :invalid_login | :invalid_password | :inactive
+      def login: [U] (*String credentials) ?{ (T, (failure_reason | nil | untyped)) -> U } -> (T | nil | U)
 
       def logged_in?: () -> bool
       def current_user: () -> (T | nil)
--- a/gems/sorcery/0.16/sorcery.rbs
+++ b/gems/sorcery/0.16/sorcery.rbs
@@ -9,7 +9,7 @@ module Sorcery
 
   module Controller
     module InstanceMethods[T]
-      def require_login: () -> (nil | void)
+      def require_login: () -> void
       def login: (*String credentials) ?{ (T, nil) -> untyped } -> (T | nil | untyped)
 
       def logged_in?: () -> bool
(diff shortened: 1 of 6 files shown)


gems/sorcery/0.16/sorcery.rbs
```diff
@@ -0,0 +1,22 @@
+module Sorcery
+  module Model
+    def authenticates_with_sorcery!: () -> void
+
+    module ClassMethods[T]
+      def authenticate: (*String credentials) ?{ () -> untyped } -> (T | nil)
```
pocke, line 6: How about using `instance` type instead of `T` type parameter?

> instance denotes the type of instance of the class. class is the singleton of the class.
>
> https://github.com/ruby/rbs/blob/master/docs/syntax.md#base-types

gems/sorcery/0.16/sorcery.rbs
```diff
@@ -0,0 +1,22 @@
...
+  module Controller
+    module InstanceMethods[T]
+      def require_login: () -> (nil | void)
+      def login: (*String credentials) ?{ (T, nil) -> untyped } -> (T | nil | untyped)
+
+      def logged_in?: () -> bool
+      def current_user: () -> (T | nil)
+      def logout: () -> (nil | void)
+      def auto_login: (T, ?bool) -> T
+      def redirect_back_or_to: (String url, ?Hash[(String | Symbol), String] flash_hash) -> (nil | void)
```
pocke, line 19: ditto for `nil | void`

gems/sorcery/0.16/sorcery.rbs
```diff
@@ -0,0 +1,22 @@
...
+
+  module Controller
+    module InstanceMethods[T]
+      def require_login: () -> (nil | void)
+      def login: (*String credentials) ?{ (T, nil) -> untyped } -> (T | nil | untyped)
+
+      def logged_in?: () -> bool
+      def current_user: () -> (T | nil)
+      def logout: () -> (nil | void)
+      def auto_login: (T, ?bool) -> T
```
pocke, line 18: How about adding a parameter name? Especially, users cannot imagine the second parameter's meaning from `?bool`.

pocke (changes requested):
Thanks for your pull request! I left comments so please check them.

gems/sorcery/0.16/sorcery.rbs
```diff
@@ -0,0 +1,22 @@
...
+
+    module ClassMethods[T]
+      def authenticate: (*String credentials) ?{ () -> untyped } -> (T | nil)
+    end
+  end
+
+  module Controller
+    module InstanceMethods[T]
+      def require_login: () -> (nil | void)
+      def login: (*String credentials) ?{ (T, nil) -> untyped } -> (T | nil | untyped)
```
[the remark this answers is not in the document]
m11o: Thank you for your advice.
I was missing the `yield` call. Line 62 in #login method always receive nil as the second argument.
`failure_reason` returns symbol (:invalid_login, :invalid_password or :inactive) or untyped. In this case, what is appropriate types?
I tried to think type below. How about do you think?

```ruby
module Controller
    module InstanceMethods[T]
      ...
      def login: (*String credentials) ?{ (T, (:invalid_login | :invalid_password | :inactive | nil | untyped)) -> untyped } -> (T | nil | untyped)
```

ref: https://github.com/Sorcery/sorcery/blob/4485701dff638bfbca428bec94dcf4de5215dd29/lib/sorcery/controller.rb#L62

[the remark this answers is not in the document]
m11o: Defining alias type is great idea!
I defined alias type as failure_reason because it can be used for login and authenticate methods. 
d6ea5a0dd6671cb15acaccb66311f52ce80ab7ad 1eac1244fbf74c6a83f7e519eca9ff8dcbc8e2b8

pocke (approved):
Looks good, thanks!

PR #207, opened 2022-09-13
[activerecord] More accurate type #find_in_batches

sadgb:
Find in batches returns an Array 

 MasterWallet.find_in_batches{|x| puts x.class, x.first.class }
  MasterWallet Load (6.3ms)  SELECT "master_wallets".* FROM "master_wallets" ORDER BY "master_wallets"."id" ASC LIMIT $1  [["LIMIT", 1000]]
Array
MasterWallet


pocke (commented):
Thanks!

pocke (approved):

PR #209, opened 2022-09-19
Change block to optional for `ActiveRecord::Relation#find_or_initialize_by`

fugakkbn:
`ActiveRecord::Relation#find_or_initialize_by` type definition requires a block.
If we try to run without blocks, we will get the following error:

```
[error] The method cannot be called without a block
│
└     admin_user = User.find_or_initialize_by(role: admin)
```

However, if we look at the method definitions, we will see that they can accept blocks.

```ruby
def find_or_initialize_by(attributes, &block)
  find_by(attributes) || new(attributes, &block)
end
```
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L231-L233

So this PR change to make the block optional.


pocke (approved):
Looks good to me, thanks!

PR #211, opened 2022-09-21
Add some type definitions to Faraday

craftscat:
Add some type definitions to [Faraday](https://github.com/lostisland/faraday).
Since I am still new to writing RBS, I only added methods that I currently use and do not have a type definition.

* `Faraday.default_adopter`
* `Faraday::Response#success?`
* `Faraday::Response#header`

Could you please check this PR?


pocke (approved):
Looks good, thanks!

Issue #210, opened 2022-09-20
Update RBS for Rack


ParadoxV5:
I am currently tinkering with Rack and RBS, using this collection for Rack’s RBS. Unfortunately, it is quite lacklustre. It looks like it was partially prepared by @mame and TypeProf through #36 and never touched since.

I’ve inquired about Rack’s opinion on developing RBS themselves through rack/rack#1967. In summary, their general consensus is that RBS’s benefits aren’t significant enough to justify the labour of coercing Rack’s duck types to RBS’s static typing.

In that discussion, I highlighted:

> However, it has possible mistakes and undocumented fields. Examples from [`Rack::Response`](https://rubydoc.info/gems/rack/2.2.2/Rack/Response) alone:
> 
> * The attribute `Rack::Response#body` should probably follow [the `SPEC` § The Body](https://github.com/rack/rack/blob/v2.2.2/SPEC.rdoc#the-body-), but the RBS restricts it to one or an Array of `Rack::Lint` or `nil`.
> * I’ve been working around the above with `Rack::Response.[]`, whose args are untyped in the RBS.
> * `Rack::Response#each` is also missing RBS for `&block`.

While I am open to helping maintain this RBS document, I’m not affiliated with either Rack or RBS and would like your opinions on this first. Rack’s YARDocs and `Rack::Lint` are describing acceptable types (static or duck) equal to or better than the RBS, and a complete RBS for Rack doesn’t provide a unique convenience (yet) perhaps besides pocke/rbs_rails#174. **[TL;DR] Should we polish this RBS doc and get it up-to-date, or is it abandoned?**

Things to note:

* [ ] Rack contributor @jeremyevans described the major hindrance in writing RBS for Rack in the discussion mentioned above:

  > RBS is designed to make you think in forms of concrete types, but Rack is designed for duck typing, so there is always going to be an impedance mismatch, until RBS supports a duck-typing based approach (e.g. the object can be any type, but must respond to these methods).

* [ ] I also mentioned in that discussion:

  > FYI, there are tools such as [AaronC81/sord](https://github.com/AaronC81/sord) that can export YARDocs’ types to RBS. I haven't played with any myself and don’t know if they suffice.

* [ ] Rack just got a new major version recently, but the priority should probably be finishing the v2 RBS first if we don’t forsake it.

ParadoxV5:
[...]

PR #215, opened 2022-09-29
Update Rack RBS to 2.2.4

ParadoxV5:
This PR updates `rack` RBS to Rack 2.2.4 by incorporating [diffs between 2.2.2 and 2.2.4](https://github.com/rack/rack/compare/v2.2.2...2.2.4).

This PR directly `git mv` the 2.2.2 RBS and `_test` files to the 2.2.4 directory. Because Rack’s RBS has untyped entries and potentially-incorrect types (#210), I determined it is not suitable for production (yet?) and did not keep the 2.2.2 directory. This PR does *not* address the aforementioned content issues.

I ran `steep check` using Bundler (to constraint Rack’s version to 2.2.4, `Gemfile`/`.lock` not checked in) and `No type error detected. 🧋`


pocke:
Could you move the directory from `rack/2.2.2/` to `rack/2.2.4/`? 


Basically, the version should be `MAJOR.MINOR`, but without a patch version. Patch versions don't have interface changes in most cases, so it is not necessary that we make other directories for each patch version.

https://github.com/ruby/gem_rbs_collection/blob/dc7059ea0cdf67f486ca9cea2b01d53f04724652/docs/CONTRIBUTING.md#generate-a-boilerplate

ParadoxV5:
> Could you move the directory from `rack/2.2.2/` to `rack/2.2.4/`?

Aren’t they already moved, @pocke? Or did you mean to move to `rack/2.2`?

> Patch versions don't have interface changes in most cases

That is correct. IIRC, there were no method changes between 2.2.2 and 2.2.4, although there were minor differences in the exposed internals that one *could* (but unlikely *would*) use.

pocke:
> Aren’t they already moved, @pocke? Or did you mean to move to rack/2.2?


Ah, sorry. I meant moving to `rack/2.2` 🙈 Could you rename the directory?

ParadoxV5:
👌

pocke (approved):
Thanks!

PR #221, opened 2022-10-03
Support URI on HTTP verb methods of Faraday

yykamei:
Faraday's methods which are the namesake of HTTP verbs also receive an instance of `URI` for the first positional argument. In addition to them, `Request#path=` support `URI`, too.

These changes follow the upstream behavior and make `gems/faraday/2.5` load `uri`. Also, they update `test.rb` to test Faraday signatures more comprehensively, such as calling a block on `get`.

By the way, I read [Write manifest.yaml](https://github.com/ruby/gem_rbs_collection/blob/79fdb8770f7073f0fd07d3440d53eae66bbe4e99/docs/CONTRIBUTING.md#write-manifestyaml), and I thought `manifest.yaml` should be added in this PR too because `uri` is not explicitly included in the [gemspec](https://github.com/lostisland/faraday/blob/fe6e71bbe0d03134f79fad3a2a095799265da6dd/faraday.gemspec) (Should I open a PR on Faraday to add it to the file?).

Anyway, thank you for maintaining such a great repository.

### DIFF
--- a/gems/faraday/2.5/_scripts/test
+++ b/gems/faraday/2.5/_scripts/test
@@ -13,7 +13,7 @@ RBS_DIR=$(cd $(dirname $0)/..; pwd)
 # Set REPO_DIR variable to validate RBS files added to the corresponding folder
 REPO_DIR=$(cd $(dirname $0)/../../..; pwd)
 # Validate RBS files, using the bundler environment present
-bundle exec rbs --repo $REPO_DIR -r faraday:2.5 validate --silent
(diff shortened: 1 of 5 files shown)


pocke (approved):
👍

pocke:
> Should I open a PR on Faraday to add it to the file?

No, this PR is enough. We need to put `manifest.yaml` in the same place as `.rbs` files. In this case, Faraday's RBS files are in this repository, so `manifest.yaml` should be in this repository also.
If we move the `.rbs` files into https://github.com/lostisland/faraday in the future, we also need to move `manifest.yaml` file.

PR #224, opened 2022-10-04
Fix backslash escaping for the boilerplate

pocke:
The `$IFS` should be escaped `\n\t`, but it was actually raw `\n\t` accidently. This PR fixes the boilerplate and existing `_scripts/test` scripts.


pocke:
@ksss @takahashim This PR changes gems that you maintain, but I'll merge this PR before your approvals because this change doesn't affect `.rbs` files.
If you have any problems, feel free to ask me 🙌

PR #225, opened 2022-10-05
Remove `#url=` and add `#url` on Faraday

yykamei:
As far as I see, Faraday doesn't seem to have `#url=`. Instead, `#url` exists in `Faraday::Request`.

https://github.com/lostisland/faraday/blob/4024f4d4029e090e533c68c3f933c7509280592d/lib/faraday/request.rb#L71-L85

Besides, the second positional argument seems to require an instance of `Hash` or `nil`, which can also be omitted, so I updated it as well.


pocke (approved):
👍

PR #226, opened 2022-10-06
Fix `ActiveRecord::Base.after_create` to be callable with block

fugakkbn:
Calling `ActiveRecord::Base.after_create` with a block will result in the following error

```
[error] The method cannot be called with a block
│ Diagnostic ID: Ruby::UnexpectedBlockGiven
│
└   after_create do
                  ~~
```

In some cases, `ActiveRecord::Base.after_create` is called with a block.
The following is an example of call to `after_create` with block in an ActiveRecord repository test.

```ruby
class CommentWithAfterCreateUpdate < Comment
  after_create do
    update(body: "bar")
  end
end
```
https://github.com/rails/rails/blob/6ebd134a9afb1870f62ff74b96c7d92a94feca6d/activerecord/test/models/comment.rb#L98-L100

The definition location Docs also has such a description.
```ruby
##
# :method: after_create
#
# :call-seq: after_create(*args, &block)
```
https://github.com/rails/rails/blob/18707ab17fa492eb25ad2e8f9818a320dc20b823/activerecord/lib/active_record/callbacks.rb#L372-L377

Therefore, this PR modifies the type definition of `ActiveRecord::Base.after_create` so that it can be called with a block.


pocke:
Why don't you update the other callback definitions, such as `after_update` and `before_create`? They are defined by the same method so I guess we can apply the same change to them.

fugakkbn:
> Why don't you update the other callback definitions, such as after_update and before_create?

Thanks for your suggestion. The callback methods defined with `after_create` has also been fixed to accept blocks.

gems/activerecord/6.0/activerecord.rbs
```diff
@@ -52,26 +52,26 @@ module ActiveRecord
     def self.enum: (Hash[Symbol, untyped]) -> void
 
     # callbacks
-    def self.after_commit: (*untyped) -> void
+    def self.after_commit: (*untyped) ?{ () -> untyped} -> void
     def self.after_create_commit: (*untyped) -> void | ...
     def self.after_update_commit: (*untyped) -> void | ...
     def self.after_destroy_commit: (*untyped) -> void | ...
     def self.after_save_commit: (*untyped) -> void | ...
```
fugakkbn, line 59: [Question]
The methods in lines 56-59 are defined in the `ActiveRecord::Transactions` module.
https://github.com/rails/rails/blob/4962e03683a6a16627d8f130bd6b767e79963918/activerecord/lib/active_record/transactions.rb#L231-L266

These methods also accept blocks, but should they be rewritten to type definitions similar to other callbacks?
Sorry, I am not sure about the `| ...` part, so I may be asking a question that is off the mark.

```suggestion
    def self.after_create_commit: (*untyped) ?{ () -> untyped} -> void
    def self.after_update_commit: (*untyped) ?{ () -> untyped} -> void
    def self.after_destroy_commit: (*untyped) ?{ () -> untyped} -> void
    def self.after_save_commit: (*untyped) ?{ () -> untyped} -> void
```

pocke: > These methods also accept blocks, but should they be rewritten to type definitions similar to other callbacks?

The overloads are added since https://github.com/ruby/gem_rbs_collection/pull/138.

They work correctly, so I'll merge this PR without changes for after_create_commit. If you'd like to refactor them, feel free to do that. 

---

`...` is a syntax for overloading. It is not well described in the documentation, so I opened a PR to describe it. https://github.com/ruby/rbs/pull/1120

pocke (approved):
👍

PR #231, opened 2022-10-08
Add type definitions for holiday-jp gem

kotaokubo:
Add type definitions for [holiday-jp](https://github.com/holiday-jp/holiday_jp-ruby) gem.

The type definitions to be added are those generated by the rbs prototype rb command.
```
$ rbs prototype rb gems/holiday_jp/0.8/_src/lib/**/*.rb >> gems/holiday_jp/0.8/holiday_jp-generated.rbs
```

Could you please check this PR?


gems/holiday_jp/0.8/_test/test.rb
```diff
@@ -0,0 +1,6 @@
+require "holiday_jp"
+
+holidays = HolidayJp.between(Date.new(2010, 9, 14), Date.new(2010, 9, 21)) # => [#<HolidayJp::Holiday>]
+holidays.first.name # => 敬老の日
+holidays.first.date # => Mon, 20 Sep 2010
```
pocke, line 5: They just test that `untyped` accepts all methods. Because `holidays` is `::Array[Holiday] | ::Array[untyped]`, then `holidays.first` is `Holiday? | untyped?`, which is the same as `untyped`.
Could you make them more meaningful by updating the type definition or the tests?

gems/holiday_jp/0.8/holiday_jp.rbs
```diff
@@ -0,0 +1,43 @@
...
+  def self.holiday?: (untyped date) -> bool
+
+  def self.holidays: () -> untyped
+end
+
+module HolidayJp
+  class Holiday
+    WDAY_NAMES: ::Array[::String]
+
+    EN_HOLIDAY_NAMES: ::Hash[::String, "New Year's Day" | "Coming of Age Day" | "National Foundation Day" | "Vernal Equinox Day" | "Constitution Memorial Day" | "Greenery Day" | "Children's Day" | "Marine Day" | "Mountain Day" | "Respect for the Aged Day" | "Autumnal Equinox Day" | "Health and Sports Day" | "Sports Day" | "National Culture Day" | "Labor Thanksgiving Day" | "Emperor's Birthday" | "Showa Day" | "Holiday in lieu" | "Citizen's Holiday" | "The Ceremony of the Enthronement of His Majesty th Emperor (at the Seiden)" | "The Funeral Ceremony of Emperor Showa." | "The Rite of Wedding of HIH Crown Prince Naruhito"]
```
pocke, line 13: These String literals are not useful as type definitions. How about replacing them with `String`?

pocke (changes requested):
Thanks for your PR! I left some comments, so could you check them?

gems/holiday_jp/0.8/_test/test.rb
kotaokubo: @pocke 
Thank you for your comment.
I am new to rbs and am having a little trouble.

The `Holidayjp.between` method returns either a class of `HolidayJp::Holiday` or an **empty array**.
So if I run the `steep check` as `holidays.first.name` in my test, I get the following error
```
test.rb:4:15: [error] Type `(::HolidayJp::Holiday | nil)` does not have method `name`
│
└ holidays.first.name # => 敬老の日
```
As a solution, if I use the test safe navigation operator, the `steep check` succeeds, is this the right way?
```
holidays = HolidayJp.between(Date.new(2010, 9, 14), holidays = HolidayJp.between(Date.new(2010, 9, 14), Date.new(2010, 9, 21)) # => [#<HolidayJp::Holiday>]
holidays.first&.name
holidays.first&.date
```

pocke: > As a solution, if I use the test safe navigation operator, the steep check succeeds, is this the right way?


Yes, it is one of the options to write the test.
You can also use `or raise` idiom, or `Array#[]`.

```ruby
holidays = ...
holiday = holidays.first or raise
holiday.name
```

```ruby
holidays = ...
holiday = holidays[0] # Array#[] is not nil-able.
holiday.name
```


Which way is better depends on the situation.

kotaokubo: Thank you for your comment.
I will try to write a test using `Array#[]`.
Could you please review it?

kotaokubo:
@pocke 
Thank you for your comment.
I have corrected the points you pointed out.
Please review again🙏

gems/holiday_jp/0.8/holiday_jp.rbs
```diff
@@ -0,0 +1,43 @@
...
+
+    def wday_name: () -> ::String
+
+    alias week wday_name
+  end
+end
+
+module HolidayJp
+  class Holidays
+    attr_accessor holidays: untyped
```
yykamei, line 31: As far as I can see, `Holidays#holidays` seems to [be initialized as an instance of `Hash`](https://github.com/holiday-jp/holiday_jp-ruby/blob/ef8f0370b1119dca348f8dc19ec0a7154f500d26/lib/holiday_jp/holidays.rb#L8). Can we declare it as `Hash[untyped, untyped]`?

```suggestion
    attr_accessor holidays: Hash[untyped, untyped]
```

kotaokubo: @yykamei  
You are right!. I will correct it as such.
Thank you for comment.

.gitmodules
```diff
@@ -121,3 +121,6 @@
 [submodule "gems/binding_of_caller/1.0/_src"]
 	path = gems/binding_of_caller/1.0/_src
 	url = https://github.com/banister/binding_of_caller.git
+[submodule "gems/holiday_jp/0.8.1/_src"]
```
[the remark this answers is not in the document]
kotaokubo: Thank you for your comment.
I'll make the right version.

PR #233, opened 2022-10-09
Fixed to allow omission of blocks in`ActiveModel::Validations::ClassMethods#validate`

akito-fujisaki:
**Background**

When calling `ActiveModel::Validations.validate` as described below., the block could not be omitted.

```
class Comment
  include ActiveModel::Validations

  validate :must_be_friends

  def must_be_friends
    errors.add(:base, 'Must be friends to leave a comment') unless commenter.friend_of?(commentee)
  end
 end
```

**Solution**

Fixed to allow omission of blocks in`ActiveModel::Validations::ClassMethods#validate`

### DIFF
--- a/gems/activemodel/6.0/activemodel-generated.rbs
+++ b/gems/activemodel/6.0/activemodel-generated.rbs
@@ -3924,7 +3924,7 @@ module ActiveModel
       #
       # NOTE: Calling +validate+ multiple times on the same method will overwrite previous definitions.
       #
-      def validate: (*untyped args) ?{ () -> untyped } -> untyped
+      def validate: (*untyped args) ?{ (untyped) -> untyped } -> untyped
 
       # List all validators that are being used to validate the model using
       # +validates_with+ method.
--- a/gems/activemodel/6.0/activemodel-generated.rbs
+++ b/gems/activemodel/6.0/activemodel-generated.rbs
@@ -3924,7 +3924,7 @@ module ActiveModel
       #
       # NOTE: Calling +validate+ multiple times on the same method will overwrite previous definitions.
       #
-      def validate: (*untyped args) { () -> untyped } -> untyped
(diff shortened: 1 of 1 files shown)


gems/activemodel/6.0/activemodel-generated.rbs
```diff
@@ -3924,7 +3924,7 @@ module ActiveModel
       #
       # NOTE: Calling +validate+ multiple times on the same method will overwrite previous definitions.
       #
-      def validate: (*untyped args) { () -> untyped } -> untyped
+      def validate: (*untyped args) ?{ () -> untyped } -> untyped
```
pocke, line 3927: I have a question. Is `validate(:method_name) do ... end` valid?
If it's valid, your PR has no problem.
If it's invalid, I think we should use overloading method like the following.

```rbs
def validate: (*untyped args) -> untyped
            | () { () -> untyped } -> untyped
```

akito-fujisaki: Thanks for the review!!
I prepared and tested the following model.
Using overloading method was not needed.

```
class Comment
  include ActiveModel::Validations

  validate :must_be_friends do
    errors.add(:base, :invalid) 
  end

  def must_be_friends
    errors.add(:base, 'Must be friends to leave a comment') unless commenter.friend_of?(commentee)
  end
end
```

However, I encountered another problem.
Steep seems to recognize `#errors` as a singleton method of `Cooment`.

```
 [error] Type `singleton(::Comment)` does not have method `errors`
```

So I fixed it to allow passing block arguments and solved this problem with https://github.com/ruby/gem_rbs_collection/pull/233/commits/9d723258391bb1ce4d16c4025da9a6b012e4792a.

pocke (approved):
Thanks!

PR #238, opened 2022-10-11
Add RBS for faker gem

4geru:
This PR adds rbs files of [faker](https://github.com/faker-ruby/faker) gem.

### DIFF
--- a/gems/faker/2.23/_test/test.rb
+++ b/gems/faker/2.23/_test/test.rb
@@ -1,0 +1,18 @@
+# Write Ruby code to test the RBS.
+# It is type checked by `steep check` command.
+
+require "faker"
+
+Faker::Name.name + 'string'
+Faker::Internet.email
+Faker::Name.unique == Faker::Name
+Faker::Name.unique.clear
+Faker::UniqueGenerator.clear
+Faker::Lorem.unique.exclude :string, [number: 6], %w[azerty wxcvbn]
+
+Faker::Config.random = Random.new(42)
+Faker::Company.bs + 'string'
+Faker::Config.random = nil
+
+Faker::Config.locale = 'es'
+Faker::Config.locale = :es
--- a/gems/faker/2.23/faker.rbs
+++ b/gems/faker/2.23/faker.rbs
@@ -8,7 +8,7 @@ module Faker
   class UniqueGenerator
     def self.clear: () -> {}
     def clear: () -> {}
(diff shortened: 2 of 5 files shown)


gems/faker/2.23/_test/test.rb
```diff
@@ -0,0 +1,18 @@
+# Write Ruby code to test the RBS.
+# It is type checked by `steep check` command.
+
+require "faker"
+
+Faker::Name.name + 'string'
```
4geru, line 6: I wrote example codes in README.md https://github.com/faker-ruby/faker#usage
I'll create another PR when merged this pull request.

gems/faker/2.23/faker.rbs
```diff
@@ -0,0 +1,33 @@
...
+
+module Faker
+  class Base
+    def self.unique: () -> UniqueGenerator
+  end
+
+  class UniqueGenerator
+    def self.clear: () -> {}
+    def clear: () -> {}
+    def exclude: (Symbol, Array[String]|Array[Hash[Symbol, untyped]], Array[String]) -> Array[String]
```
pocke, line 11: I guess `void` is more appropriate for the returned value. This method doesn't expect the user to use the returned value.

pocke (commented):
Thanks for your PR! I left a comment, could you please check it?

gems/faker/2.23/faker.rbs
4geru: I run the sample codes. and it returns Array[String].
But the user expects a void like your suggestion.

I prefer to define Array[String]. please give your opinion. I follow you.

```ruby
Faker::Lorem.unique.exclude :string, [number: 6], %w[azerty wxcvbn]
=> ["azerty", "wxcvbn"]
```

pocke (approved):
👍

gems/faker/2.23/faker.rbs
pocke: Yeah, this method actually returns an array of string. 

The returned value is a returned value of `each` method.
https://github.com/faker-ruby/faker/blob/07234831e201c853c65120e0569d9a810f18586d/lib/helpers/unique_generator.rb#L53-L58
In many cases, the returned value of `each` method is ignored. I think this case also ignores the returned value.

If we define it as `Array[String]` instead of `void`, someone may use the value as an array. But Faker doesn't expect this usage. So I think we should avoid the wrong usage by typing it as `void`.

PR #239, opened 2022-10-11
Add RequestStore type definition

nogtk:
Add type definitions for [request_store](https://github.com/steveklabnik/request_store) gem.

I added only some methods written in [RequestStore's README](https://github.com/steveklabnik/request_store/blob/master/README.md).
This is the firest PR, and I'll add other method definitions as it is merged.

Could you review this PR?

### DIFF
--- a/gems/request_store/1.5/manifest.yaml
+++ b/gems/request_store/1.5/manifest.yaml
@@ -1,7 +1,0 @@
-# manifest.yaml describes dependencies which do not appear in the gemspec.
-# If this gem includes such dependencies, comment-out the following lines and
-# declare the dependencies.
-# If all dependencies appear in the gemspec, you should remove this file.
-#
-# dependencies:
-#   - name: pathname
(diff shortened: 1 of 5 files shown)


gems/request_store/1.5/manifest.yaml
```diff
@@ -0,0 +1,7 @@
+# manifest.yaml describes dependencies which do not appear in the gemspec.
+# If this gem includes such dependencies, comment-out the following lines and
+# declare the dependencies.
+# If all dependencies appear in the gemspec, you should remove this file.
```
pocke, line 4: Could you remove this file?

pocke (changes requested):
Thanks for your PR! I just left one comment, could you please confirm that?

gems/request_store/1.5/manifest.yaml
nogtk: OK, I deleted in this commit https://github.com/ruby/gem_rbs_collection/pull/239/commits/ba20bf1598b33fad7191824e47c8bca6655595ff

nogtk:
@pocke 
Thank you for reviewing!
I have sent re-Request review. Could you please confirm the fixed commit.

pocke (approved):
Thanks!

PR #240, opened 2022-10-12
Add rbs for hashie gem

sue445:
I add rbs for [Hashie::Mash](https://github.com/hashie/hashie/blob/v5.0.0/lib/hashie/mash.rb)


pocke (approved):
Looks good, thanks!

PR #242, opened 2022-10-14
Add type definitions for moji gem

fugakkbn:
This PR adds RBS for [moji](https://github.com/gimite/moji) gem.

We will now add RBS for two methods, `Moji.han_to_zen` and `Moji.zen_to_han`.
These methods are defined at:
https://github.com/gimite/moji/blob/master/lib/moji.rb#L468-L510

These are defined as module functions.
https://github.com/gimite/moji/blob/master/lib/moji.rb#L546-L549

Therefore, RBS is also defined as singleton methods.

### DIFF
--- a/gems/moji/1.6/moji.rbs
+++ b/gems/moji/1.6/moji.rbs
@@ -1,0 +1,1 @@
+# Write the type definition here!
--- a/gems/moji/1.6/moji.rbs
+++ b/gems/moji/1.6/moji.rbs
@@ -1,5 +1,5 @@
 module Moji
-  def self?.han_to_zen: (::String str, ?untyped tp) -> String
+  def self?.han_to_zen: (::String str, ?untyped type) -> String
 
-  def self?.zen_to_han: (::String str, ?untyped tp) -> String
+  def self?.zen_to_han: (::String str, ?untyped type) -> String
 end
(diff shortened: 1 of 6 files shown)


gems/moji/1.6/moji.rbs
```diff
@@ -0,0 +1,5 @@
+module Moji
+  def self.han_to_zen: (::String str, ?untyped tp) -> String
```
pocke, line 2: As you said, these methods are defined as `module_function`s. How about using `self?` instead of `self`? `self?` represents module function feature.

https://github.com/ruby/rbs/blob/master/docs/syntax.md#method-definition

> self? method definition adds two methods: a public singleton method and a private instance method, which is equivalent to module_function in Ruby.

pocke, line 2: How about rename the second argument to `type`? It is named as `type` in the documentation, and `type` is easier to understand the meaning than `tp`.

https://github.com/gimite/moji/blob/fe5013d404a744f1c521145e17943d17385327f8/lib/moji.rb#L186

pocke (changes requested):
Thanks for your PR!
I left some comments, could you check them?

gems/moji/1.6/moji.rbs
fugakkbn: Certainly the `type` is easier to understand. I'll set the second argument to `type`.

By the way, is there a problem with using `type` as an argument name? I curious because we also use `type` in Type alias declarations.

pocke: Probably it doesn't have any problems. If I remember correctly, RBS allows keywords as an argument name. You can also quote the keyword with backticks if you get a syntax error.

pocke (approved):
Thanks!

PR #243, opened 2022-10-16
Add RBS for jwt@2.5

autopp:
This PR add RBS for [jwt@2.5](https://github.com/jwt/ruby-jwt/tree/v2.5.0).

Currently this RBS contains definition of
- `JWT.encode` and `JWT.decode` (incomplete)
- Error classes

### DIFF
--- a/gems/jwt/2.5/jwt.rbs
+++ b/gems/jwt/2.5/jwt.rbs
@@ -1,0 +1,48 @@
+module JWT
+  interface _Algo
+    def alg: () -> ::String
+    def sign: (data: ::String, signing_key: untyped) -> ::String
+    def valid_alg?: (String alg_to_validate) -> boolish
+    def verify: (data: untyped, signature: ::String, verification_key: untyped) -> boolish
+  end
+
+  def self?.encode: (untyped payload, untyped key, ?::String | _Algo algorithm, ?::Hash[::String | ::Symbol, untyped] header_fields) -> ::String
+
+  def self?.decode: (String jwt, ?untyped? key, ?boolish verify, ?::Hash[Symbol, untyped] options) ?{ (Hash[::String, untyped] header, ?untyped payload) -> untyped } -> [untyped, Hash[::String, untyped]]
+
+
+  class EncodeError < StandardError
+  end
+  class DecodeError < StandardError
+  end
+  class RequiredDependencyError < StandardError
+  end
+
+  class VerificationError < DecodeError
+  end
+  class ExpiredSignature < DecodeError
+  end
+  class IncorrectAlgorithm < DecodeError
+  end
+  class ImmatureSignature < DecodeError
+  end
+  class InvalidIssuerError < DecodeError
+  end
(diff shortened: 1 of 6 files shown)


gems/jwt/2.5/jwt.rbs
```diff
@@ -0,0 +1,48 @@
...
+  interface _Algo
+    def alg: () -> ::String
+    def sign: (data: ::String, signing_key: untyped) -> ::String
+    def valid_alg?: (String alg_to_validate) -> bool
+    def verify: (data: untyped, signature: ::String, verification_key: untyped) -> bool
+  end
+
+  def self?.encode: (untyped payload, untyped key, ::String | _Algo algorithm, ?::Hash[::String | ::Symbol, untyped] header_fields) -> ::String
+
+  def self?.decode: (String jwt, ?untyped? key, ?bool verify, ?::Hash[Symbol, untyped] options) ?{ (Hash[::String, untyped] header, ?untyped payload) -> untyped } -> [untyped, Hash[::String, untyped]]
```
pocke, line 11:
```suggestion
  def self?.decode: (String jwt, ?untyped? key, ?boolish verify, ?::Hash[Symbol, untyped] options) ?{ (Hash[::String, untyped] header, ?untyped payload) -> untyped } -> [untyped, Hash[::String, untyped]]
```

How about using `boolish` instead of `bool`?
https://github.com/ruby/rbs/blob/30e800622a2d7217f822beb2693bdff16f528193/docs/syntax.md#bool-or-boolish

`decode` method receives anything as `verify`, and the argument is used only for `if`/`unless` conditions. So I think we have no reason to strict it to `true` or `false`.

gems/jwt/2.5/jwt.rbs
```diff
@@ -0,0 +1,48 @@
+module JWT
+  interface _Algo
+    def alg: () -> ::String
+    def sign: (data: ::String, signing_key: untyped) -> ::String
+    def valid_alg?: (String alg_to_validate) -> bool
+    def verify: (data: untyped, signature: ::String, verification_key: untyped) -> bool
+  end
+
+  def self?.encode: (untyped payload, untyped key, ::String | _Algo algorithm, ?::Hash[::String | ::Symbol, untyped] header_fields) -> ::String
```
pocke, line 9:
```suggestion
  def self?.encode: (untyped payload, untyped key, ?::String | _Algo algorithm, ?::Hash[::String | ::Symbol, untyped] header_fields) -> ::String
```

`algorithm` argument looks optional.
https://github.com/jwt/ruby-jwt/blob/b1b52504a7d7819240f2bc9a50da1a79ed03df2b/lib/jwt.rb#L21

autopp: Sorry I was wrong.

autopp:
@pocke Thanks for review. I reflected your comment at [d0f8db2](https://github.com/ruby/gem_rbs_collection/pull/243/commits/d0f8db25350782575367d26e0711dfb00f00dc6e).

pocke (approved):
LGTM 👍 Thanks for your PR!

PR #245, opened 2022-10-18
Add `Faraday::VERSION` and add @yykamei to CODEOWNERS

yykamei:
I got hit with an error of `Faraday::VERSION` is missing when I run `steep check` on my gem.

`Faraday::VERSION` is defined here.

https://github.com/lostisland/faraday/blob/4024f4d4029e090e533c68c3f933c7509280592d/lib/faraday/version.rb#L4

By the way, I'm interested in maintaining Faraday signatures, so I also updated `CODEOWNERS`.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,6 +6,7 @@
 /gems/ast            @pocke
 /gems/aws-sdk-*      @ksss
 /gems/dogstatsd-ruby @VTRyo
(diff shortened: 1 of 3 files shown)


pocke (approved):
Thank you!

I've invited you to be a collaborator. Please check the email 📧

yykamei:
Thank you!

PR #246, opened 2022-10-19
Add Faraday::FlatParamsEncoder

yykamei:
Previously, Faraday users may encounter the error of `UnknownConstant` when they configure the Faraday instance with `Faraday::FlatParamsEncoder`. This module is used for assuming URI params as a string including repetitive query keys, such as `a=123&a=234&a=30` (it will be decoded as `{ "a" => ["123", "234", "30"] }`). The usage is described on the [doc](https://lostisland.github.io/faraday/usage/customize). I think the module is worth being defined in this repository.

Note `Faraday::FlatParamsEncoder` is implemented in this file.

https://github.com/lostisland/faraday/blob/36916f07596f6e1ab71b688be56e5e087a1372df/lib/faraday/encoders/flat_params_encoder.rb


pocke (approved):
Looks good, thanks!

yykamei:
Thank you! It's awesome before I ask you to review!

PR #250, opened 2022-10-24
Add `Faraday::Middleware` and `Faraday::MiddlewareRegistry`

yykamei:
I added `Faraday::Middleware` and `Faraday::MiddlewareRegistry`. These module signatures might be useful for Faraday plugin authors.


pocke (approved):
🚀

PR #251, opened 2022-10-24
New signature: binding_of_caller

ursm:
https://github.com/banister/binding_of_caller

`binding_of_caller`, which allows access to binding per call stack, by John Mair (banisterfiend).


pocke (approved):
Thanks!

PR #252, opened 2022-10-24
Unify quoting style

ursm:
How about making the quoting consistent within the generated Steepfile? I'm sorry to be so detailed, but I'm just wondering.


pocke (approved):
I do not have a strong opinion about cosmetic changes.

PR #254, opened 2022-10-30
Add rbs for spreadsheet gem

fugakkbn:
Add RBS for [spreadsheet](https://github.com/zdavatz/spreadsheet) gem.

The definition of `Spreadsheet.open` is from:
https://github.com/zdavatz/spreadsheet/blob/master/lib/spreadsheet.rb#L63-L73

The first argument `io_or_path` is a String file path as shown in the usage example. Then `File.open` is called in the else clause, passing the File instance to `Spreadsheet.open` to call. So, the type of `io_or_path` is `String | File`.


pocke (approved):
Thanks!

PR #255, opened 2022-10-30
Add business_time:0.13

ksss:
https://github.com/bokmann/business_time

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,6 +5,7 @@
 # Maintainer of individual gems (alphabetical order)
 /gems/ast               @pocke
 /gems/aws-sdk-*         @ksss
+/gems/business_time     @ksss
 /gems/dogstatsd-ruby    @VTRyo
 /gems/faraday           @yykamei
 /gems/i18n              @takahashim
--- a/gems/business_time/0.13/manifest.yaml
+++ b/gems/business_time/0.13/manifest.yaml
@@ -1,0 +1,12 @@
+dependencies:
+  - name: date
+  - name: time
+  - name: monitor
+  - name: singleton
+  - name: logger
+  - name: mutex_m
+  - name: minitest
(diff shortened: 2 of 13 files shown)


gems/business_time/0.13/manifest.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - name: time
+  - name: monitor
+  - name: singleton
+  - name: logger
+  - name: mutex_m
+  - name: minitest
+  - name: json
+  - name: benchmark
+  - name: digest
+  - name: set
```
pocke, line 12: I guess most of all dependencies are activesupport's. If so, they are unnecessary here.

ksss: I removed it because the dependency on activesupport alone will provide everything.

ksss:
I rebased by current main and add enable "untyped call" checking.

.github/CODEOWNERS
```diff
@@ -12,6 +12,14 @@
...
 /gems/rails-dom-testing @ksss
 /gems/yard              @ksss
+/gems/ast            @pocke
+/gems/aws-sdk-*      @ksss
+/gems/business_time  @ksss
+/gems/dogstatsd-ruby @VTRyo
+/gems/faraday        @yykamei
+/gems/i18n           @takahashim
+/gems/rolify         @ksss
+/gems/stackprof      @pocke
```
pocke, line 22: Could you fix this file to avoid duplication?

ksss: fixed 👍

gems/business_time/0.13/manifest.yaml
ksss: Only `set` was left.

ksss:
Updated all!

pocke (approved):
Thanks!

PR #256, opened 2022-10-31
Check untyped call automatically by test

pocke:
This PR adds a test to check "untyped call" by `steep stats` command.

# Problem

Sometimes test code includes untyped calls unexpectedly. For example:

```ruby
# rbs
def f: () -> untyped

# ruby
f.something # <- untyped call
```

In the test file, this method call is unexpected. In this case, any method calls for `f` is allowed because `f` returns untyped. So it tests nothing.

I have commented on this problem more than once. e.g. https://github.com/ruby/gem_rbs_collection/pull/222#discussion_r986491783
So I'd like to avoid this problem automatically.

# Solution

Use `steep stats`.

# Failed gems

* [x] actionpack (due to rack)
* [x] actionview (due to rack)
* [x] activejob (due to rack)
* [x] activerecord (6.0 and 6.1) (due to rack)
* [x] activesupport (due to rack)
* [ ] aws-sdk-core (untyped call)
  * Temporary disabled because I'm not the owner
* [x] delayed_job (due to rack)
* [x] i18n (untyped call)
* [x] nokogiri (untyped call)
* [ ] parallel (untyped call)
  * Disabled because it looks difficult... 
* [x] paranoia (untyepd call)
* [x] redis (untyped call)
* [ ] retryable (something wrong)
  * Temporary disabled because of Steep's bug? https://github.com/soutaro/steep/issues/666
* [x] sentry-ruby (due to rack)
* [x] sidekiq (untyped call)


yykamei (approved):
Thank you! It's so helpful.

ksss:
Am I correct in understanding that this validation only validates method calls from `untyped`, but allows methods that intentionally return untyped to calls, such as `eval`?

If so, this seems like a good approach.

Another approach would be to run the test code by ruby. What do you think?

pocke:
> Am I correct in understanding that this validation only validates method calls from untyped, but allows methods that intentionally return untyped to calls, such as eval?


Yes, you are correct.

```ruby
# rbs
def f: () -> untyped

# ruby
var = f       # It is ok
var.something # It is not ok
```

> Another approach would be to run the test code by ruby. What do you think?

It is a good idea to write more meaningful test code. I recommend keeping the test code executable as Ruby.
But executing test scripts is sometimes difficult. For instance, such as rails-related gems depend on many things. We also need to run `bundle install` to execute Ruby scripts.
So I prefer executable test scripts, but it is not enforced.






---------

By the way, could you update aws-sdk-core's test script to avoid untyped calls? Currently the test is commented out because the test cases have untyped calls. 
https://github.com/ruby/gem_rbs_collection/pull/256/files#diff-5b4cdecf62388d9e3b82a78b48a55e9ea8bfb14d10d69bade77741eb77272ca7R21-R22

ksss:
OK, Thank you for the explanation.
I'll try to fix aws-sdk-core.

PR #259, opened 2022-10-31
Add the type definition for `Faraday::Adapter`

yykamei:
`Faraday::Adapter` is a base class for Faraday adapters, such as [faraday-net_http](https://github.com/lostisland/faraday-net_http) and [faraday-httpclient](https://github.com/lostisland/faraday-httpclient). I think this new declaration might help Faraday plugin developers to know what kind of methods are available for their adapters.


pocke (approved):
Thanks!

PR #265, opened 2022-11-02
Fix untyped call of testing code in aws-sdk-core

ksss:
- `#config` is a struct whose structure changes dynamically depending on the plugin.
- `#handle_request` returns an anonymous class.


pocke (approved):
👍

PR #269, opened 2022-11-10
Add type signatures for `Faraday::Utils::Headers`

yykamei:
`Faraday::Utils::Headers` is defined here.

https://github.com/lostisland/faraday/blob/26a35fd76a83ea78e328774b4e475e60c809d868/lib/faraday/utils/headers.rb

I tried to add specific types, but I couldn't avoid `untyped` completely.


pocke (approved):
Thanks!

Issue #98, opened 2021-12-17
Redis is missing `zadd` and possibly more


wagenet:
https://rubydoc.info/gems/redis/Redis#zadd-instance_method

pocke:
Thanks for reporting the issue.

Currently this repository accepts incomplete type definitions, so they may include missing methods often.
Adding method definitions is welcome. Feel free to open a PR :raised_hands:

pocke:
I'll close this issue. `zadd` method has been added by #271, and this repository doesn't focus on complete type definitions. I think we don't need to open individual issues for missing methods, but PRs are welcome!

PR #271, opened 2022-11-19
Add Redis some method of Redis::Command::SortedSets

tatematsu-k:
I try this issue https://github.com/ruby/gem_rbs_collection/issues/98.

For now, I have added the definitions in the following range.
I'm not familiar with redis-rb, but I added a type with reference to comments and some sample code.
https://github.com/redis/redis-rb/blob/a8e00cbdac5e2eefc1872c0718d0290975b6e93b/lib/redis/commands/sorted_sets.rb#L3-L122


gems/redis/4.2/redis.rbs
```diff
@@ -198,4 +198,8 @@ class Redis
   # @return [String] `PONG`
   def ping: (?String? message) -> String
 
+  def zcard: (String|Symbol key) -> Integer
+  def zadd: (String|Symbol key, *untyped args, ?nx: bool, ?xx: bool, ?lt: bool, ?gt: bool, ?ch: bool, ?incr: bool) -> [bool, Integer, Float]
```
yykamei, line 202: XX and NX look to be mutually exclusive, according to [ZADD](https://redis.io/commands/zadd/). This is tested in redis-rb [here](https://github.com/redis/redis-rb/blob/v4.2.5/test/lint/sorted_sets.rb#L46). In my opinion, it would be better if we can understand whether we use incompatible options at the same time before running tests.

By the way, `:lt` and `:gt` are not supported in v4.2; they were introduced in v4.5.0.

Also, [the tests for sorted sets](https://github.com/redis/redis-rb/blob/v4.2.5/test/lint/sorted_sets.rb) explains the return type of `#zadd` can be changed based on its arguments. In v4.2, `#zadd` possibly returns `bool`, `Integer`, or `Float` from the test assertions:

* https://github.com/redis/redis-rb/blob/v4.2.5/test/lint/sorted_sets.rb#L33
* https://github.com/redis/redis-rb/blob/v4.2.5/test/lint/sorted_sets.rb#L54
* https://github.com/redis/redis-rb/blob/v4.2.5/test/lint/sorted_sets.rb#L102

Ideally, we should declare the method signature for each condition (e.g., if `:incr` is passed, the return type will be `Float`), but it would be challenging to follow all cases (at least to me).

Now, I want to suggest the following code, following the above. What do you think?

```suggestion
  def zadd: (_ToS key, *untyped args, ?nx: bool, ?ch: bool, ?incr: bool) -> bool | Integer | Float
          | (_ToS key, *untyped args, ?xx: bool, ?ch: bool, ?incr: bool) -> bool | Integer | Float
```

gems/redis/4.2/_test/sorted_sets.rb
```diff
@@ -0,0 +1,6 @@
+redis = Redis.new
+
+redis.zcard(:sample_key)
+redis.zadd(:sample_key, 1, 'hello world')
```
yykamei, line 4: Can we test other cases? For example, `#zadd` can receive `:ch`, so it would be better to check the use of this option like this:

```ruby
redis.zadd("foo", 1, "s1", ch: true)
redis.zadd("foo", 1, "s1", incr: true)
redis.zadd("foo", 1, "s1", nx: true)
redis.zadd("foo", [2, "s1", 3, "s2", 4, "s3"], xx: true)
```

yykamei (commented):
Hi, I added some comments, but totally it looks good 👍

gems/redis/4.2/redis.rbs
```diff
@@ -198,4 +198,8 @@ class Redis
   # @return [String] `PONG`
   def ping: (?String? message) -> String
 
+  def zcard: (String|Symbol key) -> Integer
```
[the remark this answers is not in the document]
tatematsu-k: I just found out about this interface for the first time.
It seems good.
https://github.com/ruby/rbs/blob/616b070dc4f5d483e8111a414dd9eac28321d43f/core/builtin.rbs#L13-L15

gems/redis/4.2/redis.rbs
tatematsu-k: > By the way, :lt and :gt are not supported in v4.2; they were introduced in v4.5.0.

This was my lack of attention. Thanks for comments and suggestion!

> Ideally, we should declare the method signature for each condition (e.g., if :incr is passed, the return type will be Float), but it would be challenging to follow all cases (at least to me).

I agree that there is a benefit to precise definition, but I find it difficult to maintain perfection. (PR for further improvement is welcome)
I think your suggestion is a good balance for me.

gems/redis/4.2/_test/sorted_sets.rb
tatematsu-k: I add some test case with optional keys.

gems/redis/4.2/redis.rbs
```diff
@@ -198,8 +198,9 @@ class Redis
...
   def ping: (?String? message) -> String
 
-  def zcard: (String|Symbol key) -> Integer
-  def zadd: (String|Symbol key, *untyped args, ?nx: bool, ?xx: bool, ?lt: bool, ?gt: bool, ?ch: bool, ?incr: bool) -> [bool, Integer, Float]
-  def zincrby: (String|Symbol key, Integer|Float increment, String member) -> Float
-  def zrem: (String|Symbol key, String|Array[String] member) -> [bool, Integer]
+  def zcard: (_ToS key) -> Integer
+  def zadd: (_ToS key, *untyped args, ?nx: bool, ?ch: bool, ?incr: bool) -> (bool | Integer | Float)
+          | (_ToS key, *untyped args, ?xx: bool, ?ch: bool, ?incr: bool) -> (bool | Integer | Float)
+  def zincrby: (_ToS key, Integer|Float increment, String member) -> Float
```
yykamei, line 204: `#zincrby` seems to [accept `"-inf"` and `"+inf"`](https://github.com/redis/redis-rb/blob/v4.2.5/test/lint/sorted_sets.rb#L145-L149), so how about adding these literals for `increment`?

```suggestion
  def zincrby: (_ToS key, Integer|Float|"-inf"|"+inf" increment, String member) -> Float
```

gems/redis/4.2/_test/sorted_sets.rb
```diff
@@ -0,0 +1,11 @@
...
+redis.zadd("string key", 1, 'hello world', nx: true)
+redis.zadd("string key", 2, 'hello world', xx: true)
+redis.zadd("string key", 3, 'hello world', ch: true)
+redis.zadd("string key", 4, 'hello world', incr: true)
+redis.zadd(:foo, [2, "s1", 3, "s2", 4, "s3"], xx: true)
+redis.zincrby(:foo, 1,  "bar")
```
yykamei, line 10: How about adding other cases, in addition to `Integer`?

```ruby
redis.zincrby(:foo, 1.0,  "bar")
redis.zincrby(:foo, "-inf",  "bar")
redis.zincrby(:foo, "+inf",  "bar")
```

yykamei (commented):
Thank you! Almost there 👍

gems/redis/4.2/redis.rbs
tatematsu-k: thanks suggestion!

gems/redis/4.2/_test/sorted_sets.rb
tatematsu-k: thanks!
add test case with https://github.com/ruby/gem_rbs_collection/pull/271/commits/eda2e166624ed88369a398b307ec8997cfe0b788

yykamei (approved):
Thank you!

LGTM

pocke (approved):

PR #276, opened 2022-11-29
Add `Faraday::Adapter::Test` for faraday

yykamei:
`Faraday::Adapter::Test` is intended for being used in tests. If you want to check RBS for test files, this new signature might be useful.

TODO:

- [x] Need rebase


yykamei:
I might have to do this.

https://github.com/tj-actions/changed-files/issues/782#issuecomment-1307069783

PR #277, opened 2022-11-30
Rails 7 components

soutaro:
This PR is to add some of the Rails 7 components. All of them are low qualified version, but I hope they help contributers to write more definitions when they find something is missing/incorrect.

* `activerecord/7.0` contains a copy of `activerecord/6.1` with a new method definition of `primary_abstract_class`
* `activesupport/7.0` and `activemodel/7.0` contain the copy of `*/6.0`, nothing changed since 6.0
* `actionmailer/7.0` has a few class and method definitions


pocke (approved):
👍

PR #279, opened 2022-11-30
Use gh-pr-diff(1) with --name-only to extract changes

yykamei:
Lately, I got an error like below on my pull request. We can fix this problem by following the error message, but I wonder if we don't have to use a GitHub Action to detect pull request changes.

```
changed-files-diff-sha
  Verifying git version...
  Valid git version found: (2.38.1)
  Running on a pull request event...
  Fetching remote refs...
  fatal: detected dubious ownership in repository at '/__w/gem_rbs_collection/gem_rbs_collection'
  To add an exception for this directory, call:
  
  	git config --global --add safe.directory /__w/gem_rbs_collection/gem_rbs_collection
  Error: Process completed with exit code 128.
```

I want to suggest this PR with the use of [gh-pr-diff(1)](https://cli.github.com/manual/gh_pr_diff). This command allows us to detect file changes with `--name-only` for a specified pull request. The advantages of this command are:

* always available on GitHub runner images
* actively maintained
* official tool

Note this change has unfortunately a disadvantage in that `test_on_changed_gems` won't run on `push` event (gh-pr-diff(1) only works on `pull_request` event). Perhaps, it would be good to configure the workflow to have the `workflow_dispatch` trigger to let us run arbitrary gems on the `main` branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,13 +29,14 @@ jobs:
     - name: Get changed gems
       id: changes
       run: |
-        echo 'gems<<GEMS' >> $GITHUB_OUTPUT
+        END_MARKER="$(uuidgen)"
+        echo "gems<<${END_MARKER}" >> $GITHUB_OUTPUT
         gh pr diff ${{ github.event.number }} --name-only | \
           grep '^gems/' | \
           awk -F / '{printf "%s/%s/%s\n", $1, $2, $3}' | \
           sort -u | \
           tee -a $GITHUB_OUTPUT
-        echo 'GEMS' >> $GITHUB_OUTPUT
+        echo "${END_MARKER}" >> $GITHUB_OUTPUT
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: 'Run test on changed gems'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,22 +18,29 @@ jobs:
     - name: 'Test syntax check'
       run: 'rbs parse gems/**/*.rbs'
   test_on_changed_gems:
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
-    container:
-      image: rubylang/ruby:3.1-focal
     steps:
     - uses: actions/checkout@v3
+    - uses: ruby/setup-ruby@v1
       with:
-        fetch-depth: 100
-    - name: 'Install dependencies'
-      run: 'bundle install --frozen'
-    - name: Get changed files
-      id: changed-files
-      uses: tj-actions/changed-files@v34.4.0
-    - name: 'Run test on changed gems'
+        ruby-version: '3.1'
+        bundler-cache: true
+    - name: Get changed gems
+      id: changes
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -18,22 +18,29 @@ jobs:
     - name: 'Test syntax check'
       run: 'rbs parse gems/**/*.rbs'
   test_on_changed_gems:
+    if: github.event_name == 'pull_request'
```
yykamei, line 21: As described in the PR, this job won't run on `push`.

.github/workflows/test.yml
```diff
@@ -18,22 +18,29 @@ jobs:
...
     runs-on: ubuntu-latest
-    container:
-      image: rubylang/ruby:3.1-focal
```
yykamei, line 23: I wanted to use GitHub CLI, so I decided not to use this image. Instead, I configured ruby/setup-ruby.

bin/check-untyped-call.rb
```diff
@@ -11,7 +11,7 @@ def green(str)
   "\e[32m#{str}\e[0m"
 end
 
-out, err, status = Open3.capture3('steep', 'stats', '--format=csv')
+out, err, status = Open3.capture3('bundle', 'exec', 'steep', 'stats', '--format=csv')
```
yykamei, line 14: This PR fails without `bundle exec`, so I added the prefix.

.github/workflows/test.yml
```diff
@@ -18,22 +18,29 @@ jobs:
...
+        gh pr diff ${{ github.event.number }} --name-only | \
+          grep '^gems/' | \
+          awk -F / '{printf "%s/%s/%s\n", $1, $2, $3}' | \
+          sort -u | \
+          tee -a $GITHUB_OUTPUT
+        echo 'GEMS' >> $GITHUB_OUTPUT
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: 'Run test on changed gems'
+      if: steps.changes.outputs.gems
```
yykamei, line 42: I extracted `if` conditional as [expressions](https://docs.github.com/en/actions/learn-github-actions/expressions). According to the document, an empty string will be converted to `0`, which means `false`.

.github/workflows/test.yml
```diff
@@ -18,22 +18,29 @@ jobs:
...
-      run: 'bundle install --frozen'
-    - name: Get changed files
-      id: changed-files
-      uses: tj-actions/changed-files@v34.4.0
+        ruby-version: '3.1'
+        bundler-cache: true
+    - name: Get changed gems
+      id: changes
+      run: |
+        echo 'gems<<GEMS' >> $GITHUB_OUTPUT
```
pocke, line 32: Could you use a UUID instead of `GEMS` as a delimiter? The fixed delimiter has a security risk.
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

> Warning: Make sure the delimiter you're using is randomly generated and unique for each run. For more information, see "[Understanding the risk of script injections](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections)".

I use `uuidgen` for this purpose in a private project.

```yaml
END_MARKER="$(uuidgen)"
echo "key<<${END_MARKER}
content
${END_MARKER}" >> $GITHUB_OUTPUT
```

yykamei: Thank you. I missed the point!

pocke (approved):
Thanks!

PR #282, opened 2022-12-07
Type definitions for Concurrent::Promises in concurrent-ruby:1.1

hanazuki:
This patch adds type definitions for concurrent-ruby, covering the new Concurrent::Promises API and the public interface of executors.

### DIFF
--- a/gems/concurrent-ruby/1.1/executor.rbs
+++ b/gems/concurrent-ruby/1.1/executor.rbs
@@ -1,0 +1,26 @@
+module Concurrent
+  module ExecutorService
+    interface _Task
+      def call: () -> void
+    end
+
+    def post: [A] (*A args) { (*A args) -> void } -> bool
+
+    def <<: (_Task task) -> bool
+
+    def can_overflow?: () -> bool
+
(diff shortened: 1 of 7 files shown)


gems/concurrent-ruby/1.1/executor.rbs
```diff
@@ -0,0 +1,26 @@
...
+    def <<: (_Task task) -> bool
+
+    def can_overflow?: () -> bool
+
+    def serialized?: () -> bool
+  end
+
+  type executor = ExecutorService | :io | :fast | :immediate
+
+  def self.executor: (executor executor_identifier) -> ExecutorService
```
yykamei, line 18: According to the Yard doc, `.executor` looks like returning `Executor`, not `ExecutorService`. Is it correct? `global_io_executor`, `global_fast_executor`, and `global_immediate_executor` are also returning other types.

https://github.com/ruby-concurrency/concurrent-ruby/blob/385156117c110808875e1e87d5267147f696f796/lib/concurrent-ruby/concurrent/configuration.rb#L165

pocke (approved):

PR #287, opened 2023-01-04
Some fix for HTTPClient

ksss:
The following fixes are proposed for the httpclient gem.

- Fixed some types
- Added automated test scripts
- Added CODEOWNERS


pocke (approved):
Thanks!

PR #289, opened 2023-01-08
use downcase instead of underscore for aws sdk gem name conversion

yyamanoi1222:
aws sdk gem names are not separated by underscore

ex. )
O AppRunner -> aws-sdk-apprunner
X AppRunner -> aws-sdk-app_runner


ksss (approved):
I confirmed.
https://github.com/aws/aws-sdk-ruby/blob/d49eeb6ac270937b918cde6ebab488afba66554b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/service.rb#L25-L27

Thanks!

PR #290, opened 2023-01-14
Add `Faraday::Error` and its subclasses

yykamei:
This patch introduces signatures for `Faraday::Error` and its subclasses. These classes are used in `:raise_error` internally, but sometimes they would be used in other middlewares. For example, [faraday-retry](https://github.com/lostisland/faraday-retry) has an option of `exceptions`, and its default value includes `Faraday::TimeoutError`. If users want to append other Faraday errors like `Faraday::ServerError`, this patch might be helpful.

### DIFF
--- a/gems/faraday/2.5/faraday.rbs
+++ b/gems/faraday/2.5/faraday.rbs
@@ -112,6 +112,66 @@ module Faraday
     def patch: (?String | URI url, ?untyped body, ?untyped headers) ?{ (Faraday::Request) -> void } -> Faraday::Response
   end
 
+  class Error < StandardError
+    attr_reader response: Hash[Symbol, untyped]?
+    attr_reader wrapped_exception: Exception?
+
+    type exc = Exception | Hash[Symbol, untyped] | String
+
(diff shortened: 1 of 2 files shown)


gems/faraday/2.5/faraday.rbs
```diff
@@ -112,6 +112,66 @@ module Faraday
...
+  class Error < StandardError
+    attr_reader response: Hash[Symbol, untyped]?
+    attr_reader wrapped_exception: Exception?
+
+    type exc = Exception | Hash[Symbol, untyped] | String
+
+    def initialize: (?exc? exc, ?Hash[Symbol, untyped]? response) -> void
+    def backtrace: () -> ::Array[String]?
+    def inspect: () -> ::String
+    def response_status: () -> untyped?
```
ksss, line 124: Since `untyped` includes `nil`, simply `untyped` is fine.

yykamei: Thank you! 71ee7d3

pocke (approved):
Thanks!

PR #297, opened 2023-03-02
Remove set library from business_time gem

pocke:
Due to https://github.com/ruby/rbs/issues/1250


pocke:
Yard also has `set` in its `manifest.yaml` so I tried to remove it too. But the generator introduces unexpected differences. So I gave up fixing Yard gem. cc/ @ksss

ksss:
@pocke Thanks for sharing.
I'll fix yard gem later.

PR #302, opened 2023-03-09
Add gimei type

osatoh:
Add [gimei](https://github.com/willnet/gimei) gem.

Only basic interfaces have been added to realize the sample code in the README.


pocke (approved):
Thanks!

PR #307, opened 2023-03-17
Drop `set` at manifest.yaml for yard

ksss:
Remove `set` from manifest.yaml since it is core in RBS3.
Also, the generator was working with RBS2 and has been updated to RBS3.


pocke (approved):
Thanks!

PR #310, opened 2023-04-09
activesupport Array#in_groups block turn to optional

TsubasaKawajiri:
Block was required in the definition, but in the actual  it was optional.
https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/array/grouping.rb#L62-L86

### DIFF
--- a/gems/activesupport/6.0/_test/test.rb
+++ b/gems/activesupport/6.0/_test/test.rb
@@ -14,3 +14,5 @@ nil.try('round', 2)
 nil.try(:tap) { |n| p n }
 nil.try { p 'hello' }
 nil.try { |n| p n }
+[1, 2, 3, 4].try { |n| p n.in_groups(2) }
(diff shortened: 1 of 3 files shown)


gems/activesupport/6.0/_test/test.rb
```diff
@@ -14,3 +14,5 @@
 nil.try(:tap) { |n| p n }
 nil.try { p 'hello' }
 nil.try { |n| p n }
+[1, 2, 3, 4].try { |n| p n.in_groups(2) }
```
pocke, line 17: Why do you use `try`? This test is for `in_groups`. Do you have any reason to call `try` for testing `in_groups`?

TsubasaKawajiri: oops I followed the other tests. but It is certainly not needed and will be removed.

PR #315, opened 2023-04-17
adding sigs for the google-protobuf gem

HoneyryderChuck:


pocke:
Could you check the CI failure? It fails on `rbs validate`.
I guess we can fix the failure by adding `-r LIB` for the libraries in the manifest.yaml as arguments of `rbs validate` in the `_scripts/test`.

pocke (approved):
Thank you!

PR #318, opened 2023-05-13
Add client options to root Resource#initialize

ksss:
Since `Aws::*::Resource#initialize` passes arguments to the client, all arguments of the Client class are required.


pocke (approved):
Thanks!

PR #322, opened 2023-05-20
Add RBS signatures for moneta

yykamei:
This patch adds RBS signatures for [moneta](https://github.com/moneta-rb/moneta).

Moneta is a Gem that provides a consistent interface by absorbing the differences between various types of data stores that support key/value data formats.

In this patch, only the basic interface of Moneta is added as RBS.


pocke (approved):
Thanks!

PR #324, opened 2023-05-21
Add RBS signatures for ActiveRecord::Enum#enum

m11o:
This PR adds RBS signatures and tests for ActiveRecord::Enum#enum.
Also, the enum signature in `6.0/activerecord.rbs` is removed and I created `6.0/activerecord-6.0.rbs` because its file is also loaded in 6.1 and 7.0 versions, and `ActiveRecord::Enum#enum has duplicated definitions` occurred.


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -49,7 +49,6 @@ module ActiveRecord
     def self.create!: (**untyped) -> instance
     def self.validate: (*untyped) -> void
     def self.validates: (*untyped) -> void
-    def self.enum: (Hash[Symbol, untyped]) -> void
```
m11o, line 52: This enum method is different from 6.0 and 7.0. And, this file is shared with other versions.
So, `ActiveRecord::Enum#enum has duplicated definitions` occurred when version 7.0.
Therefore, its line was removed from this file and added signature in each version.

gems/activerecord/6.0/activerecord-6.0.rbs
```diff
@@ -0,0 +1,5 @@
+module ActiveRecord
+  module Enum
+    def enum: (Hash[Symbol, (Hash[(Symbol | String), untyped] | Array[Symbol | String] | bool)] definitions) -> void
```
m11o, line 3: ActiveRecord::Enum#enum method in 6.0 version
https://github.com/rails/rails/blob/28bb76d3efc39b2ef663dfe2346f7c2621343cd6/activerecord/lib/active_record/enum.rb#L150

gems/activerecord/6.1/activerecord-6.1.rbs
```diff
@@ -5,4 +5,8 @@ module ActiveRecord
     end
     extend DelegatedType
   end
+
+  module Enum
+    def enum: (Hash[Symbol, (Hash[(Symbol | String), untyped] | Array[Symbol | String] | bool)] definitions) -> void
```
m11o, line 10: ActiveRecord::Enum#enum method in 6.1 version
https://github.com/rails/rails/blob/f09dc7c4c2e8b9375345d443c230cb8d78ad6a18/activerecord/lib/active_record/enum.rb#L158

gems/activerecord/7.0/activerecord-7.0.rbs
```diff
@@ -4,4 +4,8 @@ module ActiveRecord
       def primary_abstract_class: () -> void
     end
   end
+
+  module Enum
+    def enum: (?(String | Symbol) name, ?(Hash[(Symbol | String), untyped] | Array[Symbol | String]) values, **untyped options) -> void
```
m11o, line 9: ActiveRecord::Enum#enum method in 7.0 version
https://github.com/rails/rails/blob/c15ee6e7b5065db3c46afa1f025b8e45f443604e/activerecord/lib/active_record/enum.rb#L167

pocke, line 9: This method definition is misleading. 
This method actually accepts Rails 6 style enum definitions like the following.

```ruby
enum status: { active: 1, inactive: 2 }
```

But the type definition does not have the corresponding type definition. It looks only accepts `enum :status, { active: 1, inactive: 2 }` style enum.


So I think we should add a branch to clarify it receives the Rails 6 style definition.


---

By the way, the test passes because `**untyped` matches `statsu: {...}` keyword. I think it is not unclear.

m11o: Thank you for your review.
Exactly! So, I changed signatures to be able to use Rails 6 style enum definitions in Rails 7.

pocke (approved):
Thanks!

PR #329, opened 2023-06-03
activerecord: Fix return type of methods in _ActiveRecord_Relation_ClassMethods

tk0miya:
The return types of some class methods of AR::Base are wrong.  They should return the relation object instead of itself.


pocke (approved):
Thanks!

PR #330, opened 2023-06-03
activerecord: Update types for AR::Base and AR::Relation (partially)

tk0miya:
Added from :find to :in_batches in the below list.

refs: https://github.com/rails/rails/blob/main/activerecord/lib/active_record/querying.rb#L5-L23

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -375,9 +375,13 @@ interface _ActiveRecord_Relation_ClassMethods[Model, Relation, PrimaryKey]
   def where: (*untyped) -> Relation
   def exists?: (*untyped) -> bool
   def any?: () -> bool
+          | () { (Model) -> boolish } -> bool
   def many?: () -> bool
+           | () { (Model) -> boolish } -> bool
   def none?: () -> bool
+           | () { (Model) -> boolish } -> bool
   def one?: () -> bool
+          | () { (Model) -> boolish } -> bool
   def order: (*untyped) -> Relation
   def group: (*Symbol | String) -> untyped
   def distinct: () -> self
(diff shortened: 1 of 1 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -342,6 +376,10 @@ interface _ActiveRecord_Relation_ClassMethods[Model, Relation, PrimaryKey]
            | (*Symbol | String columns) -> Array[Array[untyped]]
   def where: (*untyped) -> Relation
   def exists?: (*untyped) -> bool
+  def any?: () -> bool
+  def many?: () -> bool
+  def none?: () -> bool
+  def one?: () -> bool
```
pocke, line 382: They can take a block.

By the way, I'm not sure they are actually necessary because `Enumerable` defines them. Aren't the definitions by `Enumerable` enough?

tk0miya: If my understanding is correct, the `ActiveRecord::Relation` class is surely including `Enumerable` module. So these methods are defined via it.
https://github.com/ruby/gem_rbs_collection/blob/main/gems/activerecord/6.0/activerecord-generated.rbs#L19531

On the other hand, the class methods of a subclass of ActiveRecord::Base is not so. So I added these methods only into `_ActiveRecord_Relation_ClassMethods`.

This is the result of `rbs method` (to types generated by rbs_rails).

```
root@f8da7af3f06e> bundle exec rbs -Isig method XXXX::ActiveRecord_Relation 'one?'
::XXXX::ActiveRecord_Relation#one?
  defined_in: ::Enumerable
  implementation: ::Enumerable
  accessibility: public
  types:
      () -> bool
    | () { (::XXXX) -> ::boolish } -> bool
```
```
root@f8da7af3f06e> bundle exec rbs -Isig method --singleton XXXX 'one?'
Cannot find method: one?
```

tk0miya: Fixed > block

pocke: Ah, I see. Thanks for your explanation!

pocke (approved):
Thanks!

PR #331, opened 2023-06-04
Add types for ActiveHash

tk0miya:
ActiveHash is a simple base class that allows you to use a ruby hash as a readonly datasource for an ActiveRecord-like model.

refs: https://github.com/active-hash/active_hash


pocke (approved):
Thanks!

PR #332, opened 2023-06-04
activerecord: Fix type of #*_joins, #includes, #eager_load and #preload

tk0miya:
The following methods can take strings, symbols and hashes at the same time.  So the current definitions should be merged into one instead of overloaded.

* ActiveRecord::Base#joins
* ActiveRecord::Base#left_joins
* ActiveRecord::Base#left_outer_joins
* ActiveRecord::Base#includes
* ActiveRecord::Base#eager_load
* ActiveRecord::Base#preload
* ActiveRecord::Relation#joins
* ActiveRecord::Relation#left_joins
* ActiveRecord::Relation#left_outer_joins
* ActiveRecord::Relation#includes
* ActiveRecord::Relation#eager_load
* ActiveRecord::Relation#preload

refs: https://github.com/rails/rails/blob/e88857bbb9d4e1dd64555c34541301870de4a45b/activerecord/lib/active_record/relation/query_methods.rb#L586


pocke:
Could you please fix the conflict?

tk0miya:
Updated :-)

pocke (approved):
Thanks!

PR #333, opened 2023-06-04
activerecord: Fix type of #merge

tk0miya:
The #merge method can take the relation object for other models:

    Account.joins(:articles).merge(Articles.where(...))


ksss:
It looks like `ActiveRecord::SpawnMethods#merge` will also accept `Array`, `Hash`, and `#to_proc`

```rb
    def merge(other, *rest)
      if other.is_a?(Array)
        records & other
      elsif other
        spawn.merge!(other, *rest)
      else
        raise ArgumentError, "invalid argument: #{other.inspect}."
      end
    end

    def merge!(other, *rest) # :nodoc:
      options = rest.extract_options!
      if other.is_a?(Hash)
        Relation::HashMerger.new(self, other, options[:rewhere]).merge
      elsif other.is_a?(Relation)
        Relation::Merger.new(self, other, options[:rewhere]).merge
      elsif other.respond_to?(:to_proc)
        instance_exec(&other)
      else
        raise ArgumentError, "#{other.inspect} is not an ActiveRecord::Relation"
      end
    end
```

https://api.rubyonrails.org/classes/ActiveRecord/SpawnMethods.html#method-i-merge

tk0miya:
Thank you for reviewing. Updated!

pocke:
It looks good but this change conflict with the default branch. Could you resolve the conflict?

tk0miya:
The conflict is resolved now.

pocke (approved):
Thanks!

PR #335, opened 2023-06-08
Add CODEOWNER for Parallel/Rainbow/Retryable gems

ybiquitous:
I previously updated the Rainbow gem and added the Parallel/Retryable gems.
- #1
- #6
- #21


pocke (approved):
Thanks for becoming the code owner!

PR #337, opened 2023-06-10
Update CONTRIBUTING.md

daiki0381:
Now it starts with an explanation about the owner of the gem.
So it is difficult to know if the subsequent steps are for gem owners or for non-gem owners.

I think the subsequent steps are for non-gem owners.
I think adding a statement like this would make it easier to understand.


pocke (approved):
Thanks for the improvement!

PR #338, opened 2023-06-10
Improve testing CI

ybiquitous:
This change mainly does two things.

1. Improve the `bin/test` script
    - add `--all` option
    - fix `--verbose` output
    - add more messages (warnings) for GitHub Actions
    - see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions
2. Run test for all gems on CI
    - add a new job
    - run the job only on the `main` branch
    - see the [actual sample on my fork](https://github.com/ybiquitous/gem_rbs_collection/actions/runs/5229039132)
      - image

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -4,30 +4,30 @@ require "pathname"
 require "optparse"
 require "open3"
 
+@all = false
 @verbose = false
 @repo_root = Pathname(__dir__) + "../gems"
 @repos = []
 
 OptionParser.new do |opts|
+  opts.banner = "Usage: bin/test [options] <gem...>"
+
   opts.on("--root=PATH", "Specify repository root") do |path|
     @repo_root = Pathname(path)
   end
   opts.on("--repo=PATH", "Add more repositories") do |path|
     @repos << Pathname(path)
   end
+  opts.on("-a", "--all", "Test all gems") do @all = true end
   opts.on("-v", "--verbose", "Print more messages") do @verbose = true end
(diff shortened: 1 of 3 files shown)


bin/test
```diff
@@ -37,40 +37,66 @@ def putv(message)
...
       "REPOS" => @repos.join(File::PATH_SEPARATOR),
       "RBS_DIR" => dir.to_s
     }
-    stdout, status = Open3.capture2(env, test_path.to_s)
-
-    putv(stdout)
+    status = Open3.popen2e(env, test_script.to_s) do |_stdin, stdout_stderr, thread|
+      putv stdout_stderr
+      thread.value
+    end
```
pocke, line 77: Just confirmation, the reasons for replacing `capture2` with `popen2e` are the following two things, right?


* To print stderr by changing `2` to `2e` method.
* To print output before the command finishes.

I agree with this change but I'd like to confirm the intention.

pocke (approved):
It looks good, but I left a comment for a confirmation.

bin/test
ybiquitous: Yes, that's right! I'm very sorry for missing the change reason. 🙇🏼

ybiquitous:
All gem tests passed! ✅
https://github.com/ruby/gem_rbs_collection/actions/runs/5240957801

@pocke Just to confirm, it'd be better to fix all warnings, right?

image

PR #339, opened 2023-06-10
CONTRIBUTING: Add how to edit rbs

daiki0381:
Added a description of how to contribute to a Gem that is already type-defined in gem_rbs_collection.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -15,6 +15,8 @@ Adding RBS files for a gem can be done with 4 steps.
 3. Write RBS files
 4. Write tests
 
+If the type is already defined in gem_rbs_collection, edit the RBS file, test file.
+
 (And you can open a pull request to get your code published.)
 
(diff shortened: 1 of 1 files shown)


pocke:
Personally, I think we do not need to describe how to contribute to the existing gem because the process is obvious with the existing documentation. I do not want to increase the documentation to keep it simple if the existing documentation is enough.
If this document confused you, could you explain your confusion? I'd like to know why we should add this section to the documentation. I'd like to update this document if it's reasonable.

daiki0381:
@pocke 
> Personally, I think we do not need to describe how to contribute to the existing gem because the process is obvious with the existing documentation. I do not want to increase the documentation to keep it simple if the existing documentation is enough.

As pocke said, it is not necessary to know how to contribute to an existing gem, as I think you can find it in the existing documentation. Thanks for your comment!

pocke:
I close this PR but thanks for your suggestion!

PR #340, opened 2023-06-12
Improve Retryable gem

ybiquitous:
## Summary

- More reliable RBS
- Remove private RBS
- Refactor RBS (unifying to one file, etc.)
- Fix test


gems/retryable/3.0/_test/Steepfile
```diff
@@ -5,7 +5,6 @@ target :test do
 
   repo_path "../../../"
   library "retryable:3.0"
-  library "forwardable"
```
ybiquitous, line 8: [note] Remove `forwardable` for public RBS.

pocke (approved):
Thanks!

PR #341, opened 2023-06-16
Import prime files from ruby/rbs

hsbt:
from https://github.com/ruby/rbs/pull/1355


pocke (approved):
👍

hsbt:
Note: We lost test case of `test/stdlib/Prime_test.rb` from `ruby/rbs`. I have no idea how import it to this repository.

PR #343, opened 2023-06-29
Run `test_all_gems` job with Dependabot PR

ybiquitous:
In addition, this change cleans up the GitHub Actions workflow a bit.

- Use `${{ ... }}` notation for consistency.
  (see https://docs.github.com/en/actions/learn-github-actions/expressions)
- Reduce repetition for Ruby versions.

Ref: https://github.com/ruby/gem_rbs_collection/pull/317#issuecomment-1612395857


pocke:
I'm against enforcing that all tests successfully finish on a gem update PR.
This change forces us to keep that all gems are valid. The maintenance cost is high, and increasing by adding new gems.

I think it is ok if the tests run on a dependabot PR but the status is green even if it contains a test failure. In this case, we can notice the test failure but we can choose whether we fix the failure or not.

ybiquitous:
> I'm against enforcing that all tests successfully finish on a gem update PR.
> This change forces us to keep that all gems are valid. The maintenance cost is high, and increasing by adding new gems.

Just to confirm, do you mind "a gem update PR" including `gems/<some_gem>/<some_version>/*` changes will run `bin/test --all`?

This PR will perform only the changed gem's test as before. 🤔

pocke:
> Just to confirm, do you mind "a gem update PR" including gems/<some_gem>/<some_version>/* changes will run bin/test --all?


No. If I understand correctly, `bin/test --all` does not run if PR contains changes for `gems/<some_gem>/<some_version>/*`, right? This behavior is not changed so I think it is no problem 👌 


My concern is a PR without `gems/<some_gem>/<some_version>/*` changes. For example, dependabot PR, README updates, and so on. I'd like to avoid a test failure for a gem becoming a merge blocker.

PR #346, opened 2023-07-03
Declare dependency from activesupport to securerandom

soutaro:
ActiveSupport implicitly depends on securerandom currently.


pocke (approved):
Thanks!

PR #353, opened 2023-07-17
switching aws s3 body type to Types::body

HoneyryderChuck:
Types::body can either be an IO or a String (the latter was missing for the type to be considered complete)


ksss:
@HoneyryderChuck Thank you for your contribution.

rbs files under `gems/aws-sdk-*` are files generated by aws-sdk-rbs-generator.
So it cannot be merged as it is.

I will fix the problem.
If you can create an issue, I can refer to it.

HoneyryderChuck:
@ksss thx for the feedback. FYI I was also planning to add signatures for aws-sdk-sns and aws-sdk-firehose (as those are the ones I have immediate interest in). Am I correct to assume that I should add them in the generator?

ksss:
Great.
aws-sdk-sns and aws-sdk-firehose are easy to add.
I'm considering a policy of adding them when requested.
I'll add them here when https://github.com/ruby/gem_rbs_collection/pull/354 is done.

ksss:
aws-sdk-sns and aws-sdk-firehose have already been added.
Please check here. https://github.com/ruby/gem_rbs_collection/issues/355

HoneyryderChuck:
Thx for the update, will close this then.

PR #354, opened 2023-07-18
Streaming input should add type StringIO and File

ksss:
ref: https://github.com/ruby/gem_rbs_collection/pull/353
partial fix: https://github.com/ruby/gem_rbs_collection/issues/355

For the `"streaming":true` input property, since aws-sdk-ruby says it supports `StringIO` and `File`.
I implemented it so that it is of that type.
In addition, I updated the aws-sdk-ruby reference to the latest version.


generators/aws-sdk-rbs-generator/lib/aws-sdk-rbs-generator/shape_dictionary/shape.rb
```diff
@@ -122,7 +122,11 @@ def to_rbs_returns
           "bool"
         when "blob"
           if streaming?
-            "::IO"
+            if input?
+              "::String | ::StringIO | ::File"
```
HoneyryderChuck, line 126: while documentation may say "StringIO or File", I believe that sockets would also work ([seems that they only need to answer to #read and #rewind](https://github.com/aws/aws-sdk-ruby/blob/bbe581191d8d0dc9be0f63be5b3656d655c5b8e0/gems/aws-sdk-core/lib/seahorse/client/http/request.rb#L51-L55). Perhaps `::String | ::StringIO | (_Reader & _Rewindable)` would be more accurate, even if more verbose?

HoneyryderChuck:
Overall LGTM 👌

pocke (approved):

PR #356, opened 2023-07-18
Add http rbs

sue445:
I add rbs for https://github.com/httprb/http

### DIFF
--- a/gems/http/5.1/_test/test.rb
+++ b/gems/http/5.1/_test/test.rb
@@ -1,6 +1,3 @@
-# Write Ruby code to test the RBS.
-# It is type checked by `steep check` command.
-
 require "http"
 
 response = HTTP::Response.new(
--- a/gems/http/5.1/_test/test.rb
+++ b/gems/http/5.1/_test/test.rb
@@ -1,0 +1,4 @@
+# Write Ruby code to test the RBS.
+# It is type checked by `steep check` command.
+
+require "http"
--- a/gems/http/5.1/_test/test.rb
+++ b/gems/http/5.1/_test/test.rb
@@ -8,3 +8,5 @@ response = HTTP::Response.new(
   version: "1.1",
   body: "Hello world!",
 )
+
+response.code == 200
--- a/gems/http/5.1/_test/test.rb
+++ b/gems/http/5.1/_test/test.rb
@@ -2,3 +2,9 @@
 # It is type checked by `steep check` command.
 
 require "http"
+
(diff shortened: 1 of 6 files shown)


gems/http/5.1/http.rbs
```diff
@@ -0,0 +1,47 @@
+# Write the type definition here!
+
+module HTTP
+  interface _Stringable
+    def to_s: () -> String
+  end
```
ybiquitous, line 6: [suggest] I'm unsure but the built-in `_ToS` interface may be available instead:

https://github.com/ruby/rbs/blob/v3.1.1/core/builtin.rbs#L13-L15

gems/http/5.1/_test/test.rb
```diff
@@ -0,0 +1,12 @@
+# Write Ruby code to test the RBS.
+# It is type checked by `steep check` command.
+
```
ybiquitous, line 3: [suggest] It's possible to remove the scaffold comment.

```suggestion
```

gems/http/5.1/http.rbs
```diff
@@ -0,0 +1,47 @@
+# Write the type definition here!
+
```
ybiquitous, line 2: [suggest] It's possible to remove the scaffold comment, too.

```suggestion
```

gems/http/5.1/http.rbs
```diff
@@ -0,0 +1,47 @@
...
+  class Response
+    attr_reader status: Integer
+    attr_reader version: String
+    attr_reader body: String
+    attr_reader request: HTTP::Request
+    attr_reader proxy_headers: Hash[untyped, untyped]
+
+    def initialize: (Hash[Symbol, untyped] opts) -> void
+
+    # defined by def_delegator
```
ybiquitous, line 39: [suggest] This internal comment is shown as a public doc on IDE, so it seems better to remove it.

```suggestion
```

ybiquitous (commented):
@sue445 Thanks for the Pull Request! I've commented on a few points of a bit concern.

gems/http/5.1/http.rbs
sue445: @ybiquitous Thank you! I fixed at 3b6d7cb

ybiquitous (commented):
LGTM! But I recommend getting other maintainers' approval.

sue445:
@pocke Could you review this patch?

pocke (approved):

PR #360, opened 2023-07-22
Add GraphQL::Dataloader::Source to graphql

greendrop:
Add `GraphQL::Dataloader::Source` to graphql gem.
Dataloader inherits from `GraphQL::Dataloader::Source`, so that is all that was added.

https://graphql-ruby.org/dataloader/sources.html
https://graphql-ruby.org/api-doc/1.12.0/GraphQL/Dataloader/Source.html


pocke (approved):
Thanks!

PR #374, opened 2023-08-03
active_hash: Add ActiveHash::Associations

tk0miya:
refs: https://github.com/active-hash/active_hash/tree/master#referencing-activerecord-objects-from-activehash

### DIFF
--- a/gems/active_hash/3.2/_test/test.rbs
+++ b/gems/active_hash/3.2/_test/test.rbs
@@ -2,3 +2,8 @@ class Colour < ActiveHash::Base
   include ActiveHash::Enum
   extend ActiveHash::Enum::Methods
 end
+
+class Pallete < ActiveHash::Base
+  include ActiveHash::Associations
+  extend ActiveHash::Associations::Methods
+end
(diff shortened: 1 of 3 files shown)


gems/active_hash/3.2/_test/test.rbs
```diff
@@ -2,3 +2,8 @@ class Colour < ActiveHash::Base
   include ActiveHash::Enum
   extend ActiveHash::Enum::Methods
 end
+
+class Pallete < ActiveHash::Base
+  include ActiveHash::Associations
+  extend ActiveHash::Associations::Methods
```
tk0miya, line 8: `ActiveHash::Associations` module calls `klass.extend(ActiveHash::Associations::Methos)` internally. So it would be nice if we could represent the behavior in RBS. Would you happen to have a better idea?

https://github.com/active-hash/active_hash/blob/4d08471385bf00bdddd9429c6315d0ea5aaff87e/lib/associations/associations.rb#L94-L97

ksss: There are two ways to generate extensions, either by hand or automatically.
RBS does not provide a solution to this problem.

Sorbet has `mixes_in_class_methods`. But it say "As a compromise".
https://sorbet.org/docs/abstract#interfaces-and-the-included-hook

tk0miya: Thanks. I understand. It would be nice to support it in the future :-)

pocke (approved):

PR #381, opened 2023-08-07
Fix init_new_gems

soutaro:
This PR fixes `bin/init_new_gems` to support gems with names longer than existing ones.

It fails if we give longer name with the following error:

```
bin/init_new_gem:142:in `*': negative argument (ArgumentError)

    insert_line += ' ' * (at_col - insert_line.size)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
        from bin/init_new_gem:142:in `block in <main>'
        from bin/init_new_gem:34:in `update'
        from bin/init_new_gem:127:in `<main>'
```

The `insert_line.size` may be bigger than any existing gem names.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,6 +2,12 @@
 # except files which have code owners.
 *                    @pocke
 
+# Misc
+/generators/aws-sdk-rbs-generator @ksss
+/generators/yard-generator @ksss
+
+# *** DON'T EDIT LINES BELOW (see bin/init_new_gem) ***
+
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -2,6 +2,12 @@
 # except files which have code owners.
 *                    @pocke
 
+# Misc
```
soutaro, line 5: Move the `# Misc` section before `/gems` section to make the updating the lines simpler.

pocke (approved):
Thanks!

PR #383, opened 2023-08-09
Add rbs for mime-types gem

greendrop:
Add rbs for [mime-types](https://github.com/mime-types/ruby-mime-types) gem.
The mime-types library provides a library and registry for information about MIME content type definitions.

- Add test from usage. ([types](https://github.com/mime-types/ruby-mime-types/blob/v3.5.0/lib/mime/types.rb), [type](https://github.com/mime-types/ruby-mime-types/blob/v3.5.0/lib/mime/type.rb))
- Add class and module definitions with `rbs prototype rb`
- Update mime-types rbs.


pocke (approved):
👍

PR #386, opened 2023-08-14
Add `--name` option for git submodule

soutaro:
`git submodule add` without `--name` option blocks adding one git repository as two sub modules, that happens when we want to work for two versions of a gem.


pocke (approved):
👍

PR #387, opened 2023-08-14
Add launchdarkly-server-sdk gem

soutaro:
https://docs.launchdarkly.com/sdk/server-side/ruby


gems/launchdarkly-server-sdk/6.4/stub.rbs
```diff
@@ -0,0 +1,3 @@
+module Observable
```
soutaro, line 1: Add empty `Observable` module, that is missing in stdlib definition in RBS 3.1.

PR #390, opened 2023-08-17
Add support for block form of ActiveSupport::Notifications.subscribe

sw-square:
Cf https://github.com/rails/rails/blob/main/activesupport/lib/active_support/notifications.rb#L212-L245 for the documentation about this


pocke:
Does this method for Rails 6 accept the same block? If so, you should edit the RBS file for Rails 6 instead. https://github.com/ruby/gem_rbs_collection/blob/a4c633634493ab7ae73219022f56acff56ab69af/gems/activesupport/6.0/activesupport-generated.rbs#L9931

pocke:
I'll close this PR because of a lack of activity.

PR #395, opened 2023-08-27
Add types for cancancan gem

tk0miya:
CanCanCan is an authorization library for Ruby and Ruby on Rails which restricts what resources a given user is allowed to access.

refs: https://github.com/CanCanCommunity/cancancan


tk0miya:
Note: cancancan gem also provides the `ActiveModel::Base#accessible_by` method. But it isn't easy to give me a type for the method. I believe the generator approach is better to support it (like rbs_rails does).
https://github.com/CanCanCommunity/cancancan#fetching-records

pocke (approved):

PR #396, opened 2023-08-28
Fix activesupport ERB definition

soutaro:
The definition causes `RBS::DuplicatedMethodDefinitionError` with `erb` type definition library, and this PR fixes the problem by:

1. Adding `erb` as `activesupport`'s dependency
2. Make the `html_escape` and `h` methods overloads

### DIFF
--- a/gems/activesupport/6.0/activesupport-generated.rbs
+++ b/gems/activesupport/6.0/activesupport-generated.rbs
@@ -5609,9 +5609,9 @@ class ERB
     #
     #   puts html_escape('is a > 0 & a < 10?')
     #   # => is a &gt; 0 &amp; a &lt; 10?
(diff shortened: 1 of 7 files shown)


gems/activesupport/6.0/activesupport-generated.rbs
```diff
@@ -5609,9 +5609,9 @@ class ERB
     #
     #   puts html_escape('is a > 0 & a < 10?')
     #   # => is a &gt; 0 &amp; a &lt; 10?
-    def self?.html_escape: (untyped s) -> untyped
+    def self?.html_escape: ...
```
soutaro, line 5612: Technically, we don't need the `def` here. Adding the `def` syntax to tell an implementation is added by activesupport.

pocke (approved):

PR #401, opened 2023-08-30
actionpack: ActionController::Base should provide .helper_method

tk0miya:
Currently `ActionController::Base` does not provide `.helper_method` because it does not extend `AbstractController::Helpers::ClassMethods`. The module was indirectly used through `AbstractController::helpers`.

So it should be added the list of related modules.

refs: https://github.com/rails/rails/blob/v6.1.7.1/actionpack/lib/action_controller/metal/helpers.rb#L60

Before:

```
root@8eadc9fcf58d> rbs -Isig method --singleton ActionController::Base helper_method
Cannot find method: helper_method
```

After:
```
root@8eadc9fcf58d> rbs -Isig method --singleton ActionController::Base helper_method
::ActionController::Base.helper_method
  defined_in: ::AbstractController::Helpers::ClassMethods
  implementation: ::AbstractController::Helpers::ClassMethods
  accessibility: public
  types:
      (*untyped meths) -> untyped
```

PR #402, opened 2023-08-31
Update dependencies for activesupport

tk0miya:
Since #396, activesupport gem has depended on erb.  But the dependencies of some gems are not updated yet.

This adds a new dependency "erb" to gems using activesupport.

refs: https://github.com/ruby/gem_rbs_collection/pull/401#issuecomment-1698911421


ksss (commented):
I approve for gem I manage.
However, many gems are managed by @pocke, so the actual approval is handed over to @pocke.

soutaro (approved):
🙏

PR #405, opened 2023-09-02
Add types for thor gem (Thor::Actions#create_file)

tk0miya:
Thor is a simple and efficient tool for building self-documenting
command line utilities.

refs: https://github.com/rails/thor

Note: This only adds types for `Thor::Action#create_file`. It's only what I needed to add. Is this acceptable? Are more method signatures needed?

This contains https://github.com/ruby/gem_rbs_collection/pull/402.

### DIFF
--- a/gems/railties/6.0/patch.rbs
+++ b/gems/railties/6.0/patch.rbs
@@ -43,21 +43,6 @@ module RDoc
   end
 end
 
-# Remove the fake RDoc type
-# if the real type of Minitest is available
-class Thor
-  class Group
-  end
-
-  module Actions
-    class CreateFile
-    end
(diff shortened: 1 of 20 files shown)


gems/railties/6.0/patch.rbs
```diff
@@ -43,21 +43,6 @@ module RDoc
   end
 end
 
-# Remove the fake RDoc type
-# if the real type of Minitest is available
```
tk0miya, line 47: I moved these definitions to the thor types.

tk0miya:
Oops. I'll fix the error tonight.

soutaro (approved):

PR #413, opened 2023-09-08
activerecord: Add modules for define a subclass of ActiveRecord::Relation

tk0miya:
Add new modules `ActiveRecord::Relation::Methods` and `ActiveRecord::
Relation::ClassMethods` to define a subclass of ActiveRecord::Relation.

So far, we have provided `_ActiveRecord_Relation` interface and
`_ActiveRecord_Relation_ClassMethods` interface for the same purpose.
However, using interfaces does not support enhancement.  For example,
users can't override the method types on the subclasses.

Therefore this starts providing modules to support the enhancement of the
subclasses.  The methods of them are the same as what the interface provides.                                                                                                                                                  

refs: https://github.com/pocke/rbs_rails/pull/254#issuecomment-1600115112

Note: Old interfaces should be keep as is for a while to keep
compatibility.

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -289,6 +289,166 @@ module ActiveRecord
   end
 end
 
+module ActiveRecord
+  class Relation
+    module Methods[Model, PrimaryKey]
+      def all: () -> self
+      def ids: () -> Array[PrimaryKey]
+      def none: () -> self
+      def pluck: (Symbol | String column) -> Array[untyped]
+              | (*Symbol | String columns) -> Array[Array[untyped]]
+      def where: (*untyped) -> self
+      def not: (*untyped) -> self
+      def exists?: (*untyped) -> bool
+      def order: (*untyped) -> self
+      def group: (*Symbol | String) -> untyped
+      def distinct: () -> self
+      def or: (self) -> self
+      def merge: (Array[untyped] | Hash[untyped, untyped] | ActiveRecord::Relation | Proc) -> self
+      def joins: (*String | Symbol | Hash[untyped, untyped]) -> self
+      def left_joins: (*String | Symbol | Hash[untyped, untyped]) -> self
+      def left_outer_joins: (*String | Symbol | Hash[untyped, untyped]) -> self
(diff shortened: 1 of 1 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -289,6 +289,167 @@ module ActiveRecord
...
+      def touch_all: (*untyped, ?time: untyped) -> untyped
+      def destroy_by: (*untyped) -> untyped
+      def delete_by: (*untyped) -> untyped
+      def each: () { (Model) -> void } -> self
+      def select: (*Symbol | String) -> self
+                | () { (Model) -> boolish } -> Array[Model]
+      def reselect: (*Symbol | String) -> self
+    end
+
+    module ClassMethods[Model, Relation, PrimaryKey]
```
ksss, line 370: Considering the future, it's a name that is likely to actually exist, so I think it would be better to have a name that is less likely to conflict.

ksss (commented):
Is it possible to avoid duplicate management with the interface side?

tk0miya:
>Is it possible to avoid duplicate management with the interface side?

It's difficult because the current release of rbs_rails depends on the interfaces. So the removal of them will cause trouble for the users. We have to keep them as is for a while.

ksss:
I have no intention of removing the interface side. I believe we should continue to maintain it.
However, as it stands, if we want to add a new method, we have to add it to both the interface and the module side.
What are your thoughts on this issue?

tk0miya:
>However, as it stands, if we want to add a new method, we have to add it to both the interface and the module side.
What are your thoughts on this issue?

Indeed. We need to discuss the policy for these components. Personally, I'll vote to call deprecation for old interfaces, and not maintain more (until removal). It's very tough to maintain both the old interface and the new module.

gems/activerecord/6.0/activerecord.rbs
tk0miya: How about `ActiveRecord::Relation::ActiveRecord_Relation_ClassMethods`? It's very redundant, but maybe safety.

pocke (approved):

PR #416, opened 2023-09-14
railties: Remove thor from manifest.yaml

tk0miya:


soutaro (approved):

soutaro:
The CI is failing because railties doesn’t have test.

PR #419, opened 2023-09-14
Add check-manifest-yaml.rb

tk0miya:
Add `check-manifest-yaml.rb` to check manifest.yaml file on each gems. It verifies

* manifest.yaml contains a "dependencies" field
* each dependency contains a "name" field
* each dependency name points a stdlib

refs: #267

PR #420, opened 2023-09-17
Add a test script to gems not having it

tk0miya:
Add a test script (a.k.a. `_scripts/test`) to gems not having it.

Note: This contains https://github.com/ruby/gem_rbs_collection/pull/419

### DIFF
--- a/gems/activestorage/6.0/_test/Steepfile
+++ b/gems/activestorage/6.0/_test/Steepfile
@@ -8,11 +8,14 @@ target :test do
   library "pathname"
   library "logger"
   library "mutex_m"
+  library "cgi"
   library "date"
+  library "erb"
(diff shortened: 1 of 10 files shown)


gems/activestorage/6.0/_test/Steepfile
```diff
@@ -8,11 +8,14 @@ target :test do
...
   library "logger"
   library "mutex_m"
+  library "cgi"
   library "date"
+  library "erb"
   library "monitor"
   library "singleton"
   library "tsort"
   library 'rack'
+  library 'thor'
```
tk0miya, line 18: I noticed the test script for activestorage:6.0 has been broken. So I also updated this at the same time.

PR #421, opened 2023-09-17
init_new_gem: Update comment on manifest.yaml

tk0miya:
It's better to describe the manifest.yaml should have a list of "stdlib" gems.


pocke:
I think manifest.yaml has to allow a gem as a dependency. https://github.com/ruby/rbs/issues/1538

pocke:
I'll close it because of a lack of activity.

PR #422, opened 2023-09-18
active_hash: Update types

tk0miya:
* Add exceptions
* Add enumerable methods to ActiveHash::Relation

refs: https://github.com/active-hash/active_hash/blob/master/lib/active_hash/relation.rb


pocke (approved):
👍

PR #427, opened 2023-09-22
actionmailer: Fix return type of ActionMailer::Base#mail

tk0miya:
The return type of `ActionMailer::Base#mail` is `Mail::Message`, not `ActionMailer::MailDelivery`.

refs:

* https://github.com/rails/rails/blob/v7.0.8/actionmailer/lib/action_mailer/base.rb#L635
* https://github.com/rails/rails/blob/v7.0.8/actionmailer/lib/action_mailer/base.rb#L858

### DIFF
--- a/gems/actionmailer/7.0/actionmailer.rbs
+++ b/gems/actionmailer/7.0/actionmailer.rbs
@@ -6,7 +6,7 @@ module ActionMailer
 
     def mail: (**untyped) ?{ (untyped) -> void } -> Mail::Message
 
-    def self.with: (**untyped) -> instance
+    def self.with: (**untyped) -> untyped
 
     def params: () -> Hash[untyped, untyped]
 
--- a/gems/actionmailer/7.0/actionmailer.rbs
+++ b/gems/actionmailer/7.0/actionmailer.rbs
@@ -4,7 +4,7 @@ module ActionMailer
 
     def self.layout: (String) -> void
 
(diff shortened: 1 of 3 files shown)


gems/actionmailer/7.0/actionmailer.rbs
```diff
@@ -4,9 +4,9 @@ module ActionMailer
 
     def self.layout: (String) -> void
 
-    def mail: (**untyped) ?{ (untyped) -> void } -> MessageDelivery
+    def mail: (**untyped) ?{ (untyped) -> void } -> Mail::Message
 
-    def self.with: (**untyped) -> instance
+    def self.with: (**untyped) -> untyped
```
tk0miya, line 9: The return type of `ActionMailer::Base.with` is `ActionMailer::Parameterized::Mailer`, not an instance of `ActionMailer::Base`.

But returning Parameterized::Mailer is not good way because it strongly depends on reflection techniques to define user defined mailer methods. Therefore it's difficult to provide correct types for the `.with` usage.

This changes the return type of the `.with` method to untyped to keep code valid.

It would be better to create a generator for ActionMailer to support correct types for `.with` method.   

refs: https://github.com/rails/rails/blob/v7.0.8/actionmailer/lib/action_mailer/parameterized.rb#L101

pocke (approved):

PR #403, opened 2023-08-31
Update activerecord callbacks

tk0miya:
This updates signatures for callback methods (ex. `#before_save`) of
ActiveRecord::Base to give the self-types to the procs and blocks.

If this is okay, I'll post another PR for actionpack (Controllers).

Note: This contains https://github.com/ruby/gem_rbs_collection/pull/402.

PR #437, opened 2023-09-30
actionpack: Update callbacks of ActionController::Base

tk0miya:
This updates signatures for callback methods (ex. #before_action) of ActiveController::Base to give the self-types to the procs and blocks.

refs: #403


pocke:
I updated the same file in https://github.com/ruby/gem_rbs_collection/pull/441 so I'll close this PR. Sorry for it 🙏

PR #445, opened 2023-10-06
Add types for chunky_png gem

tk0miya:
This library can read and write PNG files. It is written in pure Ruby for maximum portability. Let me rephrase: it does NOT require RMagick or any other memory leaking image library.

refs: https://github.com/wvanbergen/chunky_png


tk0miya:
Oops. It's my wrong. chunky_png gem has been already added to this repo. Withdrawing.

PR #450, opened 2023-10-11
activesupport: Delete a dummy definition: Object#sum

tk0miya:
It seems this definition was added to avoid the trouble on `rbs prototype rb` and refinements.  But it was gone away now because `rbs prototype rb` command ignores refinements now.

refs: https://github.com/ruby/rbs/pull/487


pocke:
The reference to #sum has been deleted in 62aadab7c8fdd3d4dbd5cd8de26fde612f47082c so we can remove this method definition.

pocke (approved):

Issue #466, opened 2023-10-28
Definition of 'parsed_response' method in httparty.rbs is wrong.


andredriem:
The method currently incorrectly indicates the gem might return a string, but looking at the [source](https://github.com/jnunemaker/httparty/blob/master/lib/httparty/parser.rb) the parsed_response method will change according to the Content-Type header.
The possible types are:
  - nil (no body in HTTP response)
  - String
  - Array[Array[untyped]]
  - Hash
  
  There is also the possibility that the user might have overridden the parser so the response might be any type the user defined as provided in the [code documentation](https://github.com/jnunemaker/httparty/blob/master/examples/custom_parsers.rb) (so this might be a common thing).  Considering this recommended use of the API the best return type should be 'untyped'.

PR #469, opened 2023-10-30
httparty: Fixed return type of HTTParty::Response#parsed_response

andredriem:
HTTParty::Response#parsed_response's return type is incorrect as described in this issue #466


andredriem:
Changed return type from untyped to Object. Also added a test call for the function

PR #473, opened 2023-11-02
globalid: Use untyped type for arguments

tk0miya:
Since RBS-3.3, void type is not allowed as an argument type.  This replaces them to untyped.


ksss (approved):
👍

pocke:
`void` indicates that the argument is not used. How about using `top` instead to clarify it?

PR #480, opened 2023-11-03
activerecord: Fix signatures of AR::Base.includes, .preload and .eager_load

tk0miya:
`ActiveRecord::Base.includes`, `.preload` and `.eager_load` can take association names in several forms; Strings, Symbols, Arrays of Strings, Hashes, and keywoard arguments.

Actually, the API document mentions such usages:

```
User.eager_load(:address, friends: [:address, :followers])
```

refs: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-eager_load

### DIFF
--- a/gems/activerecord/6.0/_test/activerecord-generated.rb
+++ b/gems/activerecord/6.0/_test/activerecord-generated.rb
@@ -10,3 +10,7 @@ class User < ActiveRecord::Base
 end
 
 _user = User.new
+
+User.eager_load(:address, friends: [:address, :followers])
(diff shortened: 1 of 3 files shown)


pocke:
Could you resolve the conflict?

BTW, `Steepfile` in `gems/` directory has not been supported since #502. Please consider another way. I guess we can use Steep's expectation file feature (it is supported).

tk0miya:
Rebased!

pocke (approved):

PR #485, opened 2023-11-18
Delegate sig from Time to TimeWithZone

ksss:
In `ActiveSupport::TimeWithZone`, many methods are common with `Time`, and there are many that can use the type definition of `Time` as is. I have updated the type definitions for some of these methods.

### DIFF
--- a/gems/activesupport/6.0/activesupport.rbs
+++ b/gems/activesupport/6.0/activesupport.rbs
@@ -76,6 +76,93 @@ module ActiveSupport
   class TimeWithZone
     include DateAndTime::Calculations
 
+    # Returns a <tt>Time</tt> instance that represents the time in +time_zone+.
+    def time: () -> ::Time
+
+    # Returns a <tt>Time</tt> instance of the simultaneous time in the UTC timezone.
+    def utc: () -> ::Time
+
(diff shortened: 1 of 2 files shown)


gems/activesupport/6.0/activesupport.rbs
```diff
@@ -76,6 +76,93 @@ module ActiveSupport
...
 
+    # Returns a <tt>Time</tt> instance that represents the time in +time_zone+.
+    def time: () -> ::Time
+
+    # Returns a <tt>Time</tt> instance of the simultaneous time in the UTC timezone.
+    def utc: () -> ::Time
+
+    # Returns a <tt>Time</tt> instance of the simultaneous time in the system timezone.
+    def localtime: (?::String utc_offset) -> ::Time
+                 | (?::Integer utc_offset) -> ::Time
```
ksss, line 87: This method does not directly use the definition from `Time`.
While `localtime` in `Time` is `(?::String utc_offset) -> ::Time`, in `TimeWithZone`, it's `alias_method :getlocal, :localtime`, so it's also necessary to consider the definition of `getlocal` in `Time`, which is `(?::Integer utc_offset) -> ::Time`.

pocke (approved):

PR #486, opened 2023-11-20
globalid: Use `top` instead of `void` to avoid error on RBS v3.3

pocke:
The globalid RBS raises an error on `rbs validate` with RBS v3.3 because RBS v3.3 does not allow `void` as an argument.

```console
$ rbs _3.2.2_ -I gems/globalid/1.1/ validate --silent
$ rbs _3.3.0_ -I gems/globalid/1.1/ validate --silent
gems/globalid/1.1/globalid.rbs:36:15...36:32: `void` type is only allowed in return type or generics parameter
gems/globalid/1.1/globalid.rbs:47:24...47:126: `void` type is only allowed in return type or generics parameter
```

By this change, `rbs validate` does not display this error.

```console
$ rbs _3.2.2_ -I gems/globalid/1.1/ validate --silent
$ rbs _3.3.0_ -I gems/globalid/1.1/ validate --silent
```

Close #473

### DIFF
--- a/gems/globalid/1.1/globalid.rbs
+++ b/gems/globalid/1.1/globalid.rbs
@@ -33,7 +33,7 @@ class GlobalID
 
   def to_param: () -> String
 
-  def as_json: (*void) -> String
+  def as_json: (*top) -> String
 
   # Delegated methods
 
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks! I missed your comment https://github.com/ruby/gem_rbs_collection/pull/473#issuecomment-1805133266. So this is very helpful :-)

PR #493, opened 2024-01-08
set sidekiq redis direct dependency

HoneyryderChuck:
After `rbs collection` on a project with sidekiq, I'm getting " Cannot find type `::Redis`" running steep. gems should be declaring their direct dependencies as well.

This is probably not the only "direct dependency" issue around, but it's enough to get the conversation started.

### DIFF
--- a/gems/sidekiq/6.3/manifest.yaml
+++ b/gems/sidekiq/6.3/manifest.yaml
@@ -1,3 +1,4 @@
 dependencies:
   - name: logger
   - name: monitor
+  - name: redis


soutaro:
> gems should be declaring their direct dependencies as well.

Dependencies between gems are not managed by rbs-collection. It assumes Bundler takes care of them and load all of the required *gems*. 
`manifest.yaml` is for *default gems*. Because the dependency to default gems are rarely declared in `Gemfile` or `.gemspec`.

The correct fix might be adding `redis-client` gem type definition in this repository.

soutaro (changes requested):
Submitting just to block someone accidentally merge this.

HoneyryderChuck:
@soutaro you're saying that sidekiq switching from "redis" to "redis-client" is the reason? ok, that was unfortunate then :/

PR #494, opened 2024-01-08
Update shoryuken

HoneyryderChuck:
Add missing types, such as middleware and main singleton methods used in default integrations.


soutaro (approved):
Thanks!

PR #495, opened 2024-01-08
Fix manifest.yaml path from CI

ksss:
When run from CI, the script was being run from the `_test` dir, so the manifest.yaml was not being checked.


soutaro (approved):
👍

PR #500, opened 2024-01-22
Create Concurrent::Map

sw-square:
This adds all of the methods for `Concurrent::Map`. It uses some metaprogramming to define superclasses/not implement methods based on which implementation of ruby it's running on, so I implemented the greatest denominator (no super class and implement all methods)


soutaro (approved):
👍

PR #502, opened 2024-01-24
Update test script

pocke:
This PR updates the test scripts system to avoid arbitrary code execution on testing. It is not a security risk for now, but this change is necessary for the new PR merging system I'm constructing.

This PR also includes the following improvements.

* Avoid duplication of `_scripts/test`
  * Previously, all `_scripts/test` files had the almost same content.
* Avoid listing dependencies by using the local source of `rbs collection`.

# How

Preivously, `_scripts/test` files in each gem had testing logic. This PR removed the script and moved the logic to `bin/test` script.

To avoid code execution, Ruby DSL files (`Steepfile` and `Gemfile`) are generated by `bin/test` script.

I also add `metadata.yaml` file to declare test dependencies. If the gem has dependencies only for the testing, the user needs to write the dependencies to `_test/matadata.yaml` file.
This file is only for testing, so it does not affect the RBS user.


bin/setup
```diff
@@ -6,3 +6,4 @@ set -v
 ROOT_DIR=$(cd $(dirname $0)/..; pwd)
 
 bundle install --gemfile=${ROOT_DIR}/Gemfile
+gem install rbs:$(bundle exec gem list rbs | ruby -e 'puts gets[/([0-9.]+)/, 1]')
```
pocke, line 9: We must install `rbs` gem globally to execute RBS without bundler context.
It is necessary for here: https://github.com/ruby/gem_rbs_collection/pull/502/files#diff-4e821476d5e4f5626c6f9fca337a47afe96a5e3daf23bccb02ad0bb109a9a407R106


`rbs collection install` command needs `Gemfile.lock` that includes the test target gems to generate `rbs_collection.lock.yaml`. But the `rbs` command should be executed in the different Bundler context than the `Gemfile.lock`. Because the `Gemfile.lock` for `rbs_collection.lock.yaml` cannot contain `rbs` gem. This `Gemfile.lock` is just generated by `bundle lock` command, so no gems are installed from the `Gemfile.lock`.





I think it is better if we can avoid the global installation. But I have no idea to avoid it 😓

soutaro: You may want to give [`bundler-compose`](https://github.com/rubygems/bundler-compose) a try?

soutaro (approved):
👍

ksss (approved):
Looks good to me.
But, I feel that it would be good to provide guidelines on how to respond to existing PR authors.

bin/setup
pocke: We've discussed `bundler-compose` and it is not appropriate for this situation.

* We require gems must not be installed, but `bundler-compose` wants to install gems
* If we use `bundler-compose`, we need to install `bundler-compose` gem globally

Another point, `rbs` gem is always installed globally because it is a bundled gem. So, the global installation does not have a large impact.

BTW, if the user installs Ruby with an OS package manager, `gem install rbs` will require `sudo` permission. This command will not work well in this case, but I think it is a rare case.

bin/test
```diff
@@ -2,22 +2,14 @@
 
 require "pathname"
 require "optparse"
-require "open3"
+require 'yaml'
 
 @all = false
 @verbose = false
-@repo_root = Pathname(__dir__) + "../gems"
-@repos = []
```
ybiquitous, line 9:
```suggestion

examples = <<~MSG
Examples:
    Test a gem
    $ bin/test gems/rack/2.2

    Test a gem with verbose messages
    $ bin/test gems/rack/2.2 -v

    Test all the gems
    $ bin/test -a
MSG
```

bin/test
```diff
@@ -2,22 +2,14 @@
...
   opts.banner = "Usage: bin/test [options] <gem...>"
 
-  opts.on("--root=PATH", "Specify repository root") do |path|
-    @repo_root = Pathname(path)
-  end
-  opts.on("--repo=PATH", "Add more repositories") do |path|
-    @repos << Pathname(path)
-  end
   opts.on("-a", "--all", "Test all gems") do @all = true end
   opts.on("-v", "--verbose", "Print more messages") do @verbose = true end
```
ybiquitous, line 14:
```suggestion
  opts.on("-v", "--verbose", "Print more messages") do @verbose = true end

  opts.on("-h", "--help", "Prints help") do
    puts "#{opts.help}\n#{examples}"
    exit
  end
```

bin/test
```diff
@@ -2,22 +2,14 @@
...
 end.parse!(ARGV)
```
ybiquitous, line 15:
```suggestion
end.tap do |opts|
  opts.parse!(ARGV)
end.tap do |opts|
  abort "#{opts.help}\n#{examples}" unless !ARGV.empty? || @all
end
```

ybiquitous (commented):
Thanks for the pull request. Great improvements!

I have a suggestion for `bin/test`: how about putting help message out to help newcomers?

E.g.

```sh-session
$ bin/test
Usage: bin/test [options] <gem...>
    -a, --all                        Test all gems
    -v, --verbose                    Print more messages
    -h, --help                       Prints help

Examples:
    Test a gem
    $ bin/test gems/rack/2.2

    Test a gem with verbose messages
    $ bin/test gems/rack/2.2 -v

    Test all the gems
    $ bin/test -a
```

pocke:
> But, I feel that it would be good to provide guidelines on how to respond to existing PR authors.


@ksss 


I'll take over existing PRs. I think I can merge most PRs. If I cannot merge a PR, I'll comment how to treat this change into the individual PR. 


> I have a suggestion for bin/test: how about putting help message out to help newcomers?



@ybiquitous 

Thanks for your suggestion. It looks good!
But I'd like to apply this change in another PR after this PR merges. The CI process for this PR is really heavy because the CI runs for all gems on this PR.
I'll create a PR with your suggestion. Thanks.

ybiquitous:
> I'll create a PR with your suggestion. Thanks.

Never mind. Glad to hear! 👍🏼

docs/CONTRIBUTING.md
```diff
@@ -34,8 +34,8 @@ This script generates an empty RBS to `gems/GEM_NAME/VERSION/GEM_NAME.rbs`, and
 
 Specify the _major_ and _minor_ version you are using would be great for most cases.
 
-We recommend adding `_scripts`, `_test`, and `_src` directories.
-We assume `_scripts/test` script runs test of the gem, `_test` directory contains files for testing, and `_src` is a git submodule for the source code of the version of the gem.
+We recommend adding `_test`, and `_src` directories.
```
ybiquitous, line 37:
```suggestion
We recommend adding `_test` and `_src` directories.
```

docs/CONTRIBUTING.md
```diff
@@ -83,14 +83,8 @@ It is important that the testing is to confirm the relationship between RBS and
 You can test your code with the following steps.
 
 1. Write a Ruby program which uses the gem code
```
ybiquitous, line 85:
```suggestion
1. Write a Ruby program which uses the gem code.
```

docs/CONTRIBUTING.md
```diff
@@ -83,14 +83,8 @@ It is important that the testing is to confirm the relationship between RBS and
...
-2. Confirm the consistency between the program and RBS
-
-#### Using Steep
-
-I have been testing with [Steep](https://github.com/soutaro/steep).
-
-Add a `Steepfile` in `_test` directory and write test scripts.
-Then you can run `steep check` to type check the test scripts.
+2. Execute `bin/test gems/GEM_NAME/VERSION` to run test test suites.
+   * It executes `rbs validate` and `steep check`.
```
ybiquitous, line 87:
```suggestion
2. Execute `bin/test gems/GEM_NAME/VERSION` to run the test suites, including `rbs validate` and `steep check`.
```

ybiquitous (approved):
I have another suggestion for the contributing guide; it includes typo fixes and an improvement for wording.

But of course, we could accept it in the following PR if this suggestion would be received.

Great job! LGTM 👍🏼

yykamei (approved):
This change looks great! I'm happy that we don't have to manage new dependencies in `_scripts/test` with  something like `-rrack`.

PR #510, opened 2024-01-29
Replace every `String | Symbol` to interned keyword

m11o:
`interned` is defined as alias for `Symbol | string` in RBS 3.3. 
ref: https://github.com/ruby/rbs/pull/1469

This PR replaced every `Symbol | string` to `interned`.
I found them by regexp: `symbol\s?\|\s?string|string\s?\|\s?symbol`.

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -81,7 +81,7 @@ module ActiveRecord
     def self.columns: () -> Array[untyped]
     def self.reflect_on_all_associations: (?Symbol) -> Array[untyped]
 
-    def will_save_change_to_attribute?: (String | Symbol attr_name, ?from: untyped, ?to: untyped) -> bool
+    def will_save_change_to_attribute?: (interned attr_name, ?from: untyped, ?to: untyped) -> bool
 
     def save!: (?context: Symbol | Array[Symbol] context, ?validate: bool, ?touch: bool) -> void
     def save: (?context: Symbol | Array[Symbol] context, ?validate: bool, ?touch: bool) -> bool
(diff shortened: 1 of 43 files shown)


ksss:
I think we should carefully consider that this change will have a positive outcome.

- `interned` is too generic a name. It should be given a name that makes sense in the context of each library.
- The intention of the creators of `interned` is not to be used publicly.
    - https://github.com/ruby/rbs/pull/1499#issuecomment-1709424448
    - " I picked interned because I doubt anyone would actually used it."

cc @sampersand

m11o:
@ksss 
I appreciate your replying.

> interned is too generic a name.

I agree with you.

> It should be given a name that makes sense in the context of each library.

As described in https://github.com/ruby/rbs/pull/1469 , `string | Symbol` is frequently used in many methods.
I think it's possible to explain the context with the method name, and I find it tedious to think of an alias every time I use the String | Symbol type.

> The intention of the creators of interned is not to be used publicly.
>   I picked interned because I doubt anyone would actually used it.

I hadn't read that comment. Thank you for sharing it 🙏 
However, after reading the comment, what I thought was that it's not about nobody using an alias named 'interned', but rather that there would be no new alias created using the word 'interned' because the word is too generic 🤔 
(Please tell me if my understanding is mistakes 😅 )

If this alias shouldn't be used publicly, it might be restricted better.

pocke:
I'm not positive about applying this change because I feel `interned` is unclear enough to indicate `String | Symbol`.


I am okay with applying this change for gems maintained by other maintainers, such as yard by ksss. Feel free to ask them if the change is acceptable.

BTW, this change will be merged more smoothly if it is separated by the gems or owners.

m11o:
@pocke 
Thank you for your advice. I will try splitting the PR by each gem. I will close this PR.

PR #533, opened 2024-03-19
Paginate calling GH API on the workflows

pocke:
Because I wasn't an admin 😂


pocke:
Ah, I need to merge this PR before trying `/merge`

PR #547, opened 2024-04-24
Specify the gem version with `.0` suffix

soutaro:
gem_rbs_collection usually assumes the directory name contains the *major* and *minor* version numbers. Generating the constraints with `~> #{version}` results in `~> 7.0`, which means testing `activerecord/7.0/_test` scripts with `activerecord/7.1` definition.

Adding `.0` generates `~> 7.0.0`, which ensures running test in `activerecord/7.0` with `activerecord/7.1` definitions.

💡  You may want to do something smarter so that directory names with patch number works correctly.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -96,7 +96,7 @@ class Runner
   end
 
   def generate_gemfile_lock
-    v = "~> #{version}"
+    v = "~> #{version}.0"
     gemfile.write(<<~RUBY)
       source 'https://rubygems.org'
       gem #{gem_name.dump}, #{v.dump}


pocke (approved):
> Most of the version specifiers, like >= 1.0, are self-explanatory. The specifier ~> has a special meaning, best shown by example. ~> 2.0.3 is identical to >= 2.0.3 and < 2.1. ~> 2.1 is identical to >= 2.1 and < 3.0. ~> 2.2.beta will match prerelease versions like 2.2.beta.12. ~> 0 is identical to >= 0.0 and < 1.0.


https://bundler.io/guides/gemfile.html


---

Looks good 👍

PR #548, opened 2024-04-28
Add active_model_serializers-0.10 gem signature

euglena1215:
https://github.com/rails-api/active_model_serializers/blob/0-10-stable

I added signatures around the methods in the range described in the README.

### DIFF
--- a/gems/active_model_serializers/0.10/_test/test.rbs
+++ b/gems/active_model_serializers/0.10/_test/test.rbs
@@ -1,29 +1,19 @@
 module Test
   class Article < ActiveModelSerializers::Model
-    module GeneratedAttributeMethods
-      attr_accessor title: String
-      attr_accessor body: String
-      attr_accessor tags: Array[Tag]
+    attr_accessor title: String
+    attr_accessor body: String
+    attr_accessor tags: Array[Tag]
 
-      def initialize: (title: String, body: String, tags: Array[Tag]) -> void
-    end
-    include GeneratedAttributeMethods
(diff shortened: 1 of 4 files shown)


gems/active_model_serializers/0.10/_test/test.rbs
```diff
@@ -0,0 +1,29 @@
...
+    include GeneratedAttributeMethods
+  end
+
+  class Tag < ActiveModelSerializers::Model
+    module GeneratedAttributeMethods
+      attr_accessor name: String
+
+      def initialize: (name: String) -> void
+    end
+    include GeneratedAttributeMethods
```
euglena1215, line 19: I thought the following implementation would work, but an error occurred.

```suggestion
    attr_accessor name: String

    def initialize: (name: String) -> void
```

```
test.rb:8:8: [error] Cannot find implementation of method `::Test::Tag#initialize`
│ Diagnostic ID: Ruby::MethodDefinitionMissing
│
└   class Tag < ActiveModelSerializers::Model
          ~~~
```

The current implementation is used because no error occurred when the module was used in reference to the rbs file generated by rbs_rails.

soutaro:
Will take a look at this PR today.

gems/active_model_serializers/0.10/_test/test.rbs
soutaro: Yeah, we can just reuse the `Model#initialize` implementation but with extra constraints with type definition. So, simply put `@dynamic` annotation works.

PR #563, opened 2024-05-26
Add rbs for mysql2-0.5

euglena1215:
https://github.com/brianmario/mysql2

### DIFF
--- a/gems/mysql2/0.5/client.rbs
+++ b/gems/mysql2/0.5/client.rbs
@@ -1,0 +1,42 @@
+module Mysql2
+  class Client
+    self.@default_query_options: untyped
+
+    @read_timeout: untyped
+
+    @query_options: untyped
+
+    attr_reader query_options: untyped
+
+    attr_reader read_timeout: untyped
+
+    def self.default_query_options: () -> untyped
+
+    def initialize: (Hash[(String | Symbol), untyped] opts) -> void
+
+    def parse_ssl_mode: (untyped mode) -> untyped
+
+    def parse_flags_array: (untyped flags, ?::Integer initial) -> untyped
+
+    # Find any default system CA paths to handle system roots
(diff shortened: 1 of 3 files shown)


gems/mysql2/0.5/client.rbs
```diff
@@ -0,0 +1,42 @@
...
+    # Find any default system CA paths to handle system roots
+    # by default if stricter validation is requested and no
+    # path is provide.
+    def find_default_ca_path: () -> untyped
+
+    # Set default program_name in performance_schema.session_connect_attrs
+    # and performance_schema.session_account_connect_attrs
+    def parse_connect_attrs: (untyped conn_attrs) -> (::Hash[untyped, untyped] | untyped)
+
+    def query: (String sql, ?::Hash[untyped, untyped] options) -> Mysql2::result
```
euglena1215, line 30: The `query` method changes the type returned by the `as` keyword argument.
If `as: :hash` or unspecified, it returns `Array[Hash]`; if `as: :array`, it returns `Array[Array]`.

If RBS could distinguish and overload literals rather than the type of the argument value, it should be able to uniquely determine the type.
```
def query: (String sql) -> Mysql2::ResultAsHash
         | (String sql, as: :hash) -> Mysql2::ResultAsHash
         | (String sql, as: :array) -> Mysql2::ResultAsArray
```

euglena1215: The above RBS syntax was found to work correctly. I was mistaken.

On the other hand, the `query` method was implemented to override the options passed in `initialize`, so the `query` method argument did not uniquely determine the type of the return value...

```
Mysql2::Client.new.query(“SELECT * FROM users”) # => Array[Hash].
Mysql2::Client.new(as: :array).query(“SELECT * FROM users”) # => maybe it should be Array[Array] 
```

PR #574, opened 2024-05-31
Update listen gem version from 3 2 to 3 9

pocke:
This PR fixes the CI problem on the default branch.


.gitmodules
```diff
@@ -4,9 +4,6 @@
 [submodule "gems/retryable/3.0/_src"]
 	path = gems/retryable/3.0/_src
 	url = https://github.com/nfedyashev/retryable.git
-[submodule "gems/listen/3.2/_src"]
```
pocke, line 7: I just removed this submodule probably we do not use

PR #578, opened 2024-06-02
activerecord: if/unless proc params can take an instance as an argument

tk0miya:
Proc objects for if/unless params to the validation and hook methods can
take an instance of the AR model as an argument.


hibariya (commented):
Looks great, thanks!

APPROVE

PR #579, opened 2024-06-02
Export from steep

ksss:
Steep has definitions for the parser, https://github.com/soutaro/steep/tree/55c58d0f8aa1fc65a8add95a5fde43f6b4375164/sig/shims
but they conflict with those in gem_rbs_collection, so only one can be used.

I believe the best approach for the community is to support and develop the ones in gem_rbs_collection.

Here are the steps I propose:

1. Stop distributing the sig directory from Steep.
2. Merge the parser definitions from Steep into gem_rbs_collection.

@soutaro @pocke 
How about you all?

### DIFF
--- a/gems/parser/3.2/parser.rbs
+++ b/gems/parser/3.2/parser.rbs
@@ -39,10 +39,490 @@ module Parser
       attr_reader location: Source::Map
       alias loc location
     end
+
+    interface _BlockNode
+      def type: () -> :block
+
+      %a{pure} def loc: () -> (Source::Map & _BlockLocation)
+    end
+
+    interface _BlockLocation
+      %a{pure} def end: () -> Source::Range
+    end
+
+    interface _DefNode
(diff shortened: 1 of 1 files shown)


pocke (approved):
I agree with this approach.

------


I think the 1st step, "Stop distributing the sig directory from Steep.", is not mandatory. This PR does not have any negative impact, so we can merge this PR before the Step change.

I'd like to wait @soutaro opinion.

pocke:
BTW I checked the contributors of the RBS files for the parser gems. There is no contributor other than Soutaro-san. So we do not need to be concerned about the license (if Soutaro-san approves it of course).
https://github.com/soutaro/steep/commits/master/sig/shims/parser

soutaro:
I didn't have an idea to export the interfaces. They are opinionated, unfinished, and not used in fact recently...
Totally 👍 for other parts.

PR #592, opened 2024-06-08
ActiveSupport Object#blank?

hibariya:


ksss:
https://github.com/rails/rails/blob/v6.0.6.1/activesupport/lib/active_support/core_ext/object/blank.rb

ksss (approved):
👍

PR #594, opened 2024-06-08
Add ActiveMerchant

hibariya:
[activemerchant](https://github.com/activemerchant/active_merchant)

For starters, added a couple of methods. I have chosen version 1.89 since it's the version I'm currently using, but let me know if there are better ways.


hibariya:
@pocke Hi, I'd like to merge this for now, but I'd like to ping you just in case since It's my first time adding a new gem here. Let me know if you find any concerns 🙏

PR #595, opened 2024-06-11
remove: ActiveModel::Serialization#send and ActiveModel::Validations#send

rhiroe:
Passing a Symbol argument to these methods will result in an error.

```
Cannot pass a value of type `::Symbol` as an argument of type `::String`
  ::Symbol <: ::String
    ::Object <: ::String
      ::BasicObject <: ::String(Ruby::ArgumentTypeMismatch)
```

Also, these methods do not exist.

https://github.com/ruby/gem_rbs_collection/commit/2e17636b572835f4b114dbb7c8942f007e71d69a
It appears to have been added by this commit, but appears to work fine without it.


ksss (commented):
APPROVE

PR #596, opened 2024-06-11
add: ActiveRecord::QueryMethods::WhereChain#missing

rhiroe:
Added `ActiveRecord::QueryMethods::WhereChain#missing` in version 6.1 or later.

Fixed to return `::ActiveRecord::QueryMethods::WhereChain` when the `where` method has no argument.

I confirmed that no errors occurred when used with [rbs-rails](https://github.com/pocke/rbs_rails).

### DIFF
--- a/gems/activerecord/6.1/activerecord-6.1.rbs
+++ b/gems/activerecord/6.1/activerecord-6.1.rbs
@@ -31,4 +31,10 @@ module ActiveRecord
   class MismatchedForeignKey < StatementInvalid
     def initialize: (?primary_key_column: untyped? primary_key_column, ?primary_key: untyped? primary_key, ?target_table: untyped? target_table, ?foreign_key: untyped? foreign_key, ?table: untyped? table, ?binds: untyped? binds, ?sql: untyped? sql, ?message: untyped? message) -> untyped
   end
+
+  module QueryMethods
+    class WhereChain
+      def missing: (*Symbol associations) -> Relation
+    end
+  end
 end
--- a/gems/activerecord/6.1/activerecord-6.1.rbs
+++ b/gems/activerecord/6.1/activerecord-6.1.rbs
@@ -33,7 +33,7 @@ module ActiveRecord
   end
 
   module QueryMethods
-    class WhereChain
+    class WhereChain[Relation]
       def missing: (*Symbol associations) -> Relation
     end
   end
--- a/gems/activerecord/6.0/activerecord-generated.rbs
+++ b/gems/activerecord/6.0/activerecord-generated.rbs
@@ -18591,40 +18591,6 @@ module ActiveRecord
 
     include ActiveModel::ForbiddenAttributesProtection
 
-    # WhereChain objects act as placeholder for queries in which #where does not have any parameter.
-    # In this case, #where must be chained with #not to return a new relation.
-    class WhereChain
-      include ActiveModel::ForbiddenAttributesProtection
-
-      def initialize: (untyped scope) -> untyped
-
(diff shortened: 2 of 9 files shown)


gems/activerecord/6.1/activerecord-6.1.rbs
```diff
@@ -31,4 +31,10 @@ module ActiveRecord
   class MismatchedForeignKey < StatementInvalid
     def initialize: (?primary_key_column: untyped? primary_key_column, ?primary_key: untyped? primary_key, ?target_table: untyped? target_table, ?foreign_key: untyped? foreign_key, ?table: untyped? table, ?binds: untyped? binds, ?sql: untyped? sql, ?message: untyped? message) -> untyped
   end
+
+  module QueryMethods
+    class WhereChain
+      def missing: (*Symbol associations) -> Relation
```
ksss, line 37: This `Relation` is `::ActiveRecord::Relation`.
So, the model information is **missing**.
Aren't you trying to get the model information from here?

How about like this?

```rbs
module QueryMethods
  class WhereChain[Relation]
    def missing: (*Symbol associations) -> Relation
  end
end
```

rhiroe: I will try it.

ksss (commented):
Almost good 👍 
Can you add testing code for `missing`?

rhiroe:
If we mimic the RBS generated by rbs-rails, we can write it like this...
```rbs
class User < ActiveRecord::Base
  extend _ActiveRecord_Relation_ClassMethods[User, ActiveRecord_Relation, Integer]

  class ActiveRecord_Relation < ::ActiveRecord::Relation
    include _ActiveRecord_Relation[User, Integer]
    include Enumerable[User]
  end

  class ActiveRecord_Associations_CollectionProxy < ::ActiveRecord::Associations::CollectionProxy
    include _ActiveRecord_Relation[User, Integer]
  end
end
```
Include or extend `_ActiveRecord_Relation_ClassMethods` and `_ActiveRecord_Relation` causes a large number of MethodDefinitionMissing errors.
```
# Type checking files:

............................................................................................................................................F................

test.rb:11:8: [error] Cannot find implementation of method `::Test::User.select`
│ Diagnostic ID: Ruby::MethodDefinitionMissing
│
└   class User < ApplicationRecord
          ~~~~

test.rb:11:8: [error] Cannot find implementation of method `::Test::User.find`
│ Diagnostic ID: Ruby::MethodDefinitionMissing
│
└   class User < ApplicationRecord
          ~~~~

......

Detected 65 problems from 1 file
No untyped calls detected. 🐾
 => Failure 🚨

# Failed gems
gems/activerecord/7.1
```
So I include or extend`::ActiveRecord::Base::ClassMethods` and  `::ActiveRecord::Relation::Methods` instead.

It looks like it was designed for that purpose.
https://github.com/ruby/gem_rbs_collection/commit/375eca7206281228ea413d754036d0362faadb61

ksss:
It is enough to show that `missing` is associated with some type, even if the code is not executable.
I do not mind if there is no code that mimics rbs_rails.

rhiroe:
@ksss Sorry, I have a question.
Is it that I am not getting Approve from you because the current test is not sufficient?

ksss:
Yes. It seems to me that there is an increase in excessively managed test code.

The test code I had in mind is as follows:

```rb
User.where.missing.to_sql
```

rhiroe:
I see. Thank you for your answer and example.

ksss (commented):
APPROVE

PR #597, opened 2024-06-11
remove: ActiveRecord::Relation::Methods#not

rhiroe:
This method does not exist.


ksss (commented):
APPROVE
The `#not` method should be in WhereChain.

PR #598, opened 2024-06-11
activerecord: CollectionProxy#build can take Hash-like object

tk0miya:
The current definition of `#build`, `#craete` and `#create!` methods of
`ActiveRecord::Associations::CollectionProxy#build` take a Hash object.
    
But they can also take Hash-like object.  Usually, Rails apps pass an
instance of `ActionController::Parameters`.  The technique is well known
as "Mass Assignment".
    
Internally, they expects the object should have `#each_pair` method.
Therefore, this changes the signature of these methods to take a
Hash-like object instead of Hash.
    
refs: https://github.com/rails/rails/blob/v7.0.0/activemodel/lib/active_model/attribute_assignment.rb#L29


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -724,10 +724,700 @@ end
 
 module ActiveRecord
   module Associations
-    class CollectionProxy
-      def build: (?::Hash[untyped, untyped] attributes) -> untyped | ...
-      def create: (?::Hash[untyped, untyped] attributes) -> untyped | ...
-      def create!: (?::Hash[untyped, untyped] attributes) -> untyped | ...
+    # Collection proxies in Active Record are middlemen between an
```
tk0miya, line 727: At first, I moved this definition from activerecord-generated.rbs to here, and merge the above definition into it.
https://github.com/ruby/gem_rbs_collection/pull/598/commits/c68dd60bee8e9691b616937d9d8be80f89cd679c

gems/activerecord/6.0/activerecord.rbs
```diff
@@ -724,10 +724,700 @@ end
...
+      #   person.pets.build([{name: 'Spook'}, {name: 'Choo-Choo'}, {name: 'Brain'}])
+      #   # => [
+      #   #      #<Pet id: nil, name: "Spook", person_id: 1>,
+      #   #      #<Pet id: nil, name: "Choo-Choo", person_id: 1>,
+      #   #      #<Pet id: nil, name: "Brain", person_id: 1>
+      #   #    ]
+      #
+      #   person.pets.size  # => 5 # size of the collection
+      #   person.pets.count # => 0 # count from database
+      def build: (?_EachPair attributes) ?{ () -> untyped } -> untyped
```
tk0miya, line 876: Second, I modified the definition of `#build`, `#create` and `#create!` to take `_EachPair`.
https://github.com/ruby/gem_rbs_collection/pull/598/commits/d0100fff01854e665dbdbc08ec888004ac005563

tk0miya, line 876: Note: I found both the type of return value and the block's argument are wrong. Users can get the built instance via the return value and the block's argument.

```
user.articles.build(params)  #=> Article
user.articles.build(params) do |article|  #=> Article
   ...
end
```

To resolve this problem, we need to change CollectionProxy to generics like `CollectionProxy[Article]`.
I think it's tough work. So I'll keep it as it is in this PR. Let's discuss it in another issue or PR.

ksss:
```diff
$ bundle exec rbs --repo gems -r activemodel diff --detail --format diff --type-name 'ActiveRecord::Associations::CollectionProxy' --before gems/activerecord/6.0 --after 598
- [::ActiveRecord::Associations::CollectionProxy public] def build: (?::Hash[untyped, untyped] attributes) -> untyped | (?::Hash[untyped, untyped] attributes) { () -> untyped } -> untyped
+ [::ActiveRecord::Associations::CollectionProxy public] def build: (?::ActiveRecord::Associations::CollectionProxy::_EachPair attributes) ?{ () -> untyped } -> untyped

- [::ActiveRecord::Associations::CollectionProxy public] def create: (?::Hash[untyped, untyped] attributes) -> untyped | (?::Hash[untyped, untyped] attributes) { () -> untyped } -> untyped
+ [::ActiveRecord::Associations::CollectionProxy public] def create: (?::ActiveRecord::Associations::CollectionProxy::_EachPair attributes) ?{ () -> untyped } -> untyped

- [::ActiveRecord::Associations::CollectionProxy public] def create!: (?::Hash[untyped, untyped] attributes) -> untyped | (?::Hash[untyped, untyped] attributes) { () -> untyped } -> untyped
+ [::ActiveRecord::Associations::CollectionProxy public] def create!: (?::ActiveRecord::Associations::CollectionProxy::_EachPair attributes) ?{ () -> untyped } -> untyped
```

gems/activerecord/6.0/activerecord.rbs
ksss: Then let's make the block argument `untyped`.

```suggestion
      def build: (?_EachPair attributes) ?{ (untyped) -> untyped } -> untyped
```

gems/activerecord/6.0/activerecord.rbs
```diff
@@ -724,10 +724,700 @@ end
...
+    #
+    # The <tt>target</tt> result set is not loaded until needed. For example,
+    #
+    #   blog.posts.count
+    #
+    # is computed directly through SQL and does not trigger by itself the
+    # instantiation of the actual post records.
+    class CollectionProxy < Relation
+      interface _EachPair
+        def each_pair: () { ([untyped, untyped]) -> untyped } -> self
```
ksss, line 753: Is `#empty?` not necessary?

https://github.com/rails/rails/blob/2ebb508cd8ee16c0bb280f91b93c01d939fcbf12/activemodel/lib/active_model/attribute_assignment.rb#L32

And `#each` ?

https://github.com/rails/rails/blob/2ebb508cd8ee16c0bb280f91b93c01d939fcbf12/activemodel/lib/active_model/attribute_assignment.rb#L41

tk0miya: I'm not sure what is the best way for such an interface. But no reason to ignore them. Okay, I'll add them soon.

tk0miya: Done. I merged https://github.com/ruby/gem_rbs_collection/pull/603 into this PR to support `ActionController::Parameters`.

gems/activerecord/6.0/activerecord.rbs
```diff
@@ -722,10 +722,702 @@ end
...
+        def each: () { ([untyped, untyped]) -> untyped } -> self
+        def empty?: () -> bool
+      end
+
+      def initialize: (untyped klass, untyped association) -> untyped
```
tk0miya, line 756: @ksss Is it better not to move these definitions unrelated to this change?

ksss: Yes. review burden will be reduced.

tk0miya: Done!

ksss (commented):
APPROVE

PR #603, opened 2024-06-12
actionpack: Add missing methods to ActionController::Parameters

tk0miya:
Some methods for ActionController::Parameters has been missing because they're defined via delegation.
    
This adds them manually.
    
refs: https://github.com/rails/rails/blob/v6.0.6.1/actionpack/lib/action_controller/metal/strong_parameters.rb#L214-L215

### DIFF
--- a/gems/actionpack/6.0/actioncontroller.rbs
+++ b/gems/actionpack/6.0/actioncontroller.rbs
@@ -156,3 +156,19 @@ module AbstractController::Callbacks::ClassMethods
   alias append_around_action around_action
   alias append_after_action after_action
 end
+
+module ActionController
+  class Parameters
+    def keys: () -> Array[untyped]
+    def key?: () -> bool
+    def has_key?: () -> bool
+    def values: () -> Array[untyped]
+    def has_value?: () -> bool
+    def empty?: () -> bool
+    def include?: (untyped) -> bool
+    def as_json: () -> String
+    def to_s: () -> String
+    def each_key: () { (untyped) -> untyped} -> Hash[untyped, untyped]
+                | () -> Enumerator[untyped, self]
+  end
+end


gems/actionpack/6.0/actioncontroller.rbs
```diff
@@ -156,3 +156,676 @@ module AbstractController::Callbacks::ClassMethods
...
+    def key?: () -> bool
+    def has_key?: () -> bool
+    def values: () -> Array[untyped]
+    def has_value?: () -> bool
+    def empty?: () -> bool
+    def include?: (untyped) -> bool
+    def as_json: () -> String
+    def to_s: () -> String
+    def each_key: () { (untyped) -> untyped} -> Hash[untyped, untyped]
+                | () -> Enumerator[untyped, self]
```
tk0miya, line 290: I added these methods.

ksss: Why didn't you use open classes?
The other changes seem redundant.

ksss:
```
$ bundle exec rbs diff --detail --format diff --type-name 'ActionController::Parameters' --before gems/actionpack/6.0 --after 603
```

```diff
- [::Kernel public] def to_s: () -> ::String
+ [::ActionController::Parameters public] def to_s: () -> ::String

- -
+ [::ActionController::Parameters public] def keys: () -> ::Array[untyped]

- -
+ [::ActionController::Parameters public] def key?: () -> bool

- -
+ [::ActionController::Parameters public] def has_key?: () -> bool

- -
+ [::ActionController::Parameters public] def values: () -> ::Array[untyped]

- -
+ [::ActionController::Parameters public] def has_value?: () -> bool

- -
+ [::ActionController::Parameters public] def empty?: () -> bool

- -
+ [::ActionController::Parameters public] def include?: (untyped) -> bool

- -
+ [::ActionController::Parameters public] def as_json: () -> ::String

- -
+ [::ActionController::Parameters public] def each_key: () { (untyped) -> untyped } -> ::Hash[untyped, untyped] | () -> ::Enumerator[untyped, self]
```

gems/actionpack/6.0/actioncontroller.rbs
tk0miya: I think this is not a patch because the original implementation is defined in the same file. So it's better to add these definitions to the same RBS file. Of course, I'm okay with using the open class way.

ksss: I am talking about moving from gems/actionpack/6.0/actionpack-generated.rbs to gems/actionpack/6.0/actioncontroller.rbs.

tk0miya: Yes, I am also talking about it. But it seems using open class is better. Okay, I update it from now on.

ksss (commented):
APPROVE

tk0miya:
@ksss Thank you for reviewing!

Issue #601, opened 2024-06-12
[Proposal] Introduce rubocop-on-rbs


ksss:
I propose introducing [rubocop-on-rbs](https://github.com/ksss/rubocop-on-rbs) to perform static analysis on RBS.

I believe that by introducing this gem, we can aim to raise the quality of RBS.

# RuboCop on RBS

[RuboCop](https://github.com/rubocop/rubocop) is one of the most popular static analysis tools for Ruby.
rubocop-on-rbs is an extension of RuboCop that enables style checks for RBS files.

# Scope

The scope of this issue is RBS files only.

# Automatic feedback

Use `github` formatter of RuboCop on GitHub Action.

```
$ rubocop --format github
```

Since feedback is displayed on the code in the PR, we can quickly notice and address the issues.

<img width="729" alt="スクリーンショット 2024-06-12 16 17 50" src="https://github.com/ruby/gem_rbs_collection/assets/935310/57431ba6-b08d-41e1-a12c-c95ea6330239">

Since it is confined to GitHub Actions, it is also secure.

# Static check only

Since we are only performing static analysis on RuboCop, there are no security issues as long as we trust the code of rubocop and rubocop-on-rbs.

# Different settings for each gem

In RuboCop, placing a `.rubocop.yml` file in a subdirectory allows you to override the settings for that directory and its subdirectories.

By utilizing this mechanism, we can change the settings for each gem individually.

https://docs.rubocop.org/rubocop/configuration.html

# On local checking

RuboCop is easy to run in a local environment, making it simple for us to verify fixes.

# Features

rubocop-on-rbs provides documentation for each feature. Please take a look at it here.

https://github.com/ksss/rubocop-on-rbs/blob/main/docs/modules/ROOT/pages/cops.adoc

# Introduction Steps

At first, I am considering applying the `RBS/Layout` and `RBS/Lint` features.
Since some of the `RBS/Style` features can be quite opinionated, I plan to introduce them later after consulting with the community.

PR #608, opened 2024-06-17
Introduce rubocop-on-rbs

ksss:
Introduce [rubocop-on-rbs](https://github.com/ksss/rubocop-on-rbs)
Implements of #601 

As an introduction, we automated the pointing out by rubocop-on-rbs to **activesupport**.

# .rubocop.yml

## Root

I have included common settings in the project root `.rubocop.yml`, but disabled all Cop.

## Each gems

I have enabled `RBS/Layout` and `RBS/Lint` only in activesupport.
With other gems, rubocop detection does not work depending on the route configuration.
This means that the owner of the gem can optionally set `.rubocop.yml` to enable the check.

# CI

By running RuboCop in CI, automatic feedback can be displayed on PRs. This frees reviewers from pointing out trivial issues and allows them to focus on essential reviews.

## Operation check

You can see that `RBS/Layout` is enabled and `RBS/Style` is not.

image

# When Updating rubocop-on-rbs

## When New Cops Are Added via gem Update

New Cops are disabled by default. There is no burden from the gem update. If you want to enable new Cops, you need to do so for each gem.

## When the Behavior of Existing Cops Changes

You need to make sure that RuboCop passes for all gems (this is assumed to be a rare case).

# When Adding a New gem

Use `init_new_gem` to add a `.rubocop.yml` file. If unnecessary, simply delete this file. This file enables most Cops by default.


pocke:
Thanks for your pull request.

Decreasing development speed is my main concern when introducing RuboCop. But probably this opt-in approach can reduce the negative impact because RuboCop is enabled only for maintainer who want to use it.


By the way, we can probably update the `init_new_gem` script to create a boilerplate of `.rubocop.yml`.

ksss:
@pocke Good idea. I updated `init_new_gem`.
And to facilitate the update of rubocop-on-rbs, I have rewritten the settings on a per-cop basis.

ksss:
@pocke 
I also thought about the operation of updating rubocop-on-rbs.
Do you have any concerns?

ksss:
@pocke 
Are you positive or negative about this change?
I would like to hear your opinion.

bin/init_new_gem
```diff
@@ -72,6 +72,71 @@ put base / 'manifest.yaml', <<~'YAML'
...
+  # You can configure RBS style of this gem.
+  # This file is used on CI. It is configured to automatically
+  # make rubocop suggestions on pull requests for this gem.
+  # If you do not like the style enforcement, you should remove this file.
+  inherit_from: ../../.rubocop.yml
+
+  ##
+  # If you want to customize the style, please consult with the gem reviewers.
+
+  RBS/Layout/CommentIndentation:
```
pocke, line 86: I think it is not a good idea to list all cops in this file, because we need to update this script manually when the gem is updated.
How about writing a link to the documentation of the cops list instead of listing all cops here?

pocke:
Sorry for the late reply. I think this approach is good for us. But I left a code comment, please check it.

bin/init_new_gem
ksss: I was concerned about the cost of increasing the version of rubocop-on-rbs.

In rubocop, new cop is supposed to warn with `Enabled: pending`. We decided to solve the problem by doing this operation in rubocop-on-rbs as well.
https://docs.rubocop.org/rubocop/versioning.html#pending-cops

ksss:
I have corrected the points raised.
Also, changes to activesupport were rescinded to simplify the PR.

.rubocop.yml
```diff
@@ -0,0 +1,30 @@
...
+    - "**/vendor/**/*"
+
+RBS:
+  Enabled: true
+RBS/Layout:
+  Enabled: false
+RBS/Lint:
+  Enabled: false
+RBS/Style:
+  Enabled: false
```
pocke, line 30: Let me confirm this configuration. Is this configuration for the `RBS` department necessary?
I guess we can simply remove these lines because `DisabledByDefault: true` is configured.

ksss: It was necessary to actually run it.

RBS/Layout is not enabled with the following settings Perhaps `DisabledByDefault: true` is in effect.

```yml
inherit_from: ../../.rubocop.yml

RBS/Layout:
  Enabled: true
```

The following settings enable `RBS/Layout`, but also enable `RBS/Lint` and `RBS/Style`.

```yml
inherit_from: ../../.rubocop.yml

RBS:
  Enabled: true
RBS/Layout:
  Enabled: true
```

pocke: I understand the problem. When `DisabledByDefault` is true, RuboCop does not enable cops under `RBS/Lint` even if the configuration enables `RBS/Lint`. But we need to enable the `RBS` department explicitly.
It sounds a bug of RuboCop, but I think it's ok to configure it in our configuration file 👍

pocke (approved):

PR #610, opened 2024-06-20
add Faraday::TooManyRequestsError

tasuku-lab280:
Added Faraday::TooManyRequestsError type added in faraday version 2.7.12
https://github.com/lostisland/faraday/releases/tag/v2.7.12

Also, the faraday version 2.7 directory was missing and has been added.


gems/faraday/2.7/faraday.rbs
```diff
@@ -0,0 +1,266 @@
...
+  end
+
+  class ConflictError < ClientError
+  end
+
+  class UnprocessableEntityError < ClientError
+  end
+
+  class TooManyRequestsError < ClientError
+  end
```
tasuku-lab280, line 156: This is a newly added type definition.

yykamei (commented):
Thank you! Looks great 👍 

As you commented [here](https://github.com/ruby/gem_rbs_collection/pull/610/files#r1646861980), the only difference between 2.5 and 2.7 is the existence of the new constant, `TooManyRequestsError`. I wonder if we can use symbolic links to avoid managing almost similar RBS files. For example, [activesupport v7.0](https://github.com/ruby/gem_rbs_collection/tree/8358df9d1582c816bc7ee2a6f91a6d744010b972/gems/activesupport/7.0) is using symbolic links to refer to v6.0 files.

From my perspective, Pros/Cons of using symbolic links can be summarized here.

Pros of using symbolic links:

* We don't have to manage similar RBS files
* We can clarify the type difference between gem versions

Cons of using symbolic links:

* If a gem drastically changes API from older versions and includes many deleted signatures, it's even worse to write patch files

In my opinion, it's better to use symbolic links for faraday v2.7 because faraday tries to stick to semantic versioning, which means there are no breaking changes among minor versions. We don't have to consider the cons of using symbolic links for faraday.

Anyway, thank you for your time to improve the faraday signatures!

tasuku-lab280:
@yykamei 

Thank you so much for your review 👍 
It is very clear!

I too felt that copying the 2.5 type file was redundant and was wondering whether to use a symbolic link or a symbolic link.

> it's better to use symbolic links for faraday v2.7 because faraday tries to stick to semantic versioning, which means there are no breaking changes among minor versions. We don't have to consider the cons of using symbolic links for faraday.

This opinion convinced me too, so I think I will change to symbolic links.
I have made the change in the following commit and would appreciate a re-review!

https://github.com/ruby/gem_rbs_collection/pull/610/commits/b636b1d21720354ab9f7ac322318f2b38b513197

yykamei (commented):
Nice 👍 

APPROVE!

PR #614, opened 2024-06-29
Allow multiple if/unless conditions in ActiveModel::Validations

aki19035vc:
Allow multiple if/unless conditions in ActiveModel::Validations as described in the Rails guide.

Reference: https://github.com/yasslab/railsguides.jp/blob/3c243cf2a78ff99a58f4646d4794810599f38a9a/guides/source/active_record_validations.md?plain=1#L1042-L1054

```ruby
class Computer < ApplicationRecord
  validates :mouse, presence: true,
                    if: [Proc.new { |c| c.market.retail? }, :desktop?],
                    unless: Proc.new { |c| c.trackpad.present? }
end
```

### DIFF
--- a/gems/activemodel/6.0/_test/test.rb
+++ b/gems/activemodel/6.0/_test/test.rb
@@ -4,7 +4,9 @@ class Person
   attr_accessor :name, :age, :email
 
   validates :name, presence: true, length: { maximum: 100 }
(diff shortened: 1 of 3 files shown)


ksss (commented):
APPROVE

PR #615, opened 2024-07-03
activemodel: add ActiveModel::Error in 7.0

aki19035vc:
**Background**

ActiveModel::Error was added in 6.1.
However, 7.0 types have not kept up with this change because it use types created from 6.0 source code.

- 6.0: https://github.com/rails/rails/blob/v6.1.7.8/activemodel/lib/active_model/errors.rb
- 7.0: https://github.com/rails/rails/blob/v7.0.8.4/activemodel/lib/active_model/errors.rb

**Solution**

- Move existing ActiveModel::Errors type definitions to activemodel-6.0.rbs
- Generate a new prototype from the 7.0 source code and write it in activemodel-7.0.rbs

**Remarks**

- If this PR is accepted, I will add the more types to ActiveModel::Errors and ActiveModel::Error.
- I separated test by module, because I may want to distinguish it from Dirty, Callbacks, etc.


ksss:
@aki19035vc Thank you for your contributing.
But, the difference was too large for me to understand the purpose of the changes.

> ActiveModel::Error was added in 6.1.

Why is a change needed for 6.0?
Why are there no changes for 6.1?

aki19035vc:
@ksss 

Thanks for the reply! I did not describe it well enough.


> Why is a change needed for 6.0?

This is because the files in the activemodel/7.0 directory are symbolic links to the 6.0 files, as shown below.

- activemodel/7.0/activemodel-generated.rbs -> activemodel/6.0/activemodel-generated.rbs
- activemodel/7.0/activemodel.rbs -> activemodel/6.0/activemodel.rbs
- activemodel/7.0/patch.rbs -> activemodel/6.0/patch.rbs

So, I think it necessary to escape the existing definitions into activemodel-6.0.rbs, if I want to reflect changes after 6.1.


> Why are there no changes for 6.1?

Because the 6.1 directory does not exist.

After creating a new directory for 6.1 and placing the files as follows,

- activemodel/6.1/activemodel-generated.rbs -> activemodel/6.0/activemodel-generated.rbs
- activemodel/6.1/activemodel.rbs -> activemodel/6.0/activemodel.rbs
- activemodel/6.1/patch.rbs -> activemodel/6.0/patch.rbs

Should I put ActiveModel::Errors and ActiveModel::Error types in the respective files below?

- activemodel/6.1/activemodel-6.1.rbs
- activemodel/7.0/activemodel-7.0.rbs

ksss:
@aki19035vc 
OK, make sense.

Whether you do or not, we should make 6.1.
Without 6.1, users using 6.1 will not be able to use `ActiveModel::Error`.

However, I am not satisfied with the current method using symbolic links.
This system tends to result in frequent file movement, which increases the review burden.
I will bring this up as a topic in the next core team meeting.

Until then, please wait a bit.
Alternatively, please narrow the scope of this PR a bit more.

aki19035vc:
@ksss 
OK! In this PR, I will not change the existing definitions, and just add the `ActiveModel::Error` type to `activemodel-7.0.rbs`.
I will further reduce the scope of this PR by not supporting 6.1.

aki19035vc:
@ksss 
I did not use `rbs prototype` and manually added the type to the public API.

Please review when you are free.

ksss (commented):
APPROVE

PR #628, opened 2024-07-22
Add headers method to Request class on Faraday

creamstew:
While attempting to add a header, I encountered the following error:

```
Type `::Faraday::Request` does not have method `headers`
```

To resolve this issue, I added a headers method to the ::Faraday::Request class.

### DIFF
--- a/gems/faraday/2.5/faraday.rbs
+++ b/gems/faraday/2.5/faraday.rbs
@@ -186,7 +186,7 @@ module Faraday
     def headers=: (untyped) -> void
     def body=: (untyped) -> void
     def options=: (untyped) -> void
-    def headers: () -> Hash[untyped, untyped]
+    def headers: () -> Utils::Headers
     def url: (String | URI path, ?::Hash[untyped, untyped] params) -> void
     def []: (untyped key) -> void
     def []=: (untyped key, untyped value) -> void
--- a/gems/faraday/2.5/faraday.rbs
+++ b/gems/faraday/2.5/faraday.rbs
@@ -186,6 +186,7 @@ module Faraday
     def headers=: (untyped) -> void
     def body=: (untyped) -> void
     def options=: (untyped) -> void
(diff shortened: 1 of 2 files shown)


gems/faraday/2.5/faraday.rbs
```diff
@@ -186,6 +186,7 @@ module Faraday
     def headers=: (untyped) -> void
     def body=: (untyped) -> void
     def options=: (untyped) -> void
+    def headers: () -> Hash[untyped, untyped]
```
yykamei, line 189: According to [the YARD doc](https://github.com/lostisland/faraday/blob/18524c6e89071fbaf0592790f0f1da7a64107959/lib/faraday/request.rb#L22), this seems to return `Utils::Headers`. As far as I see the implementation in Faraday, an instance of `Utils::Headers` is assigned to `req.headers=` [here](https://github.com/lostisland/faraday/blob/18524c6e89071fbaf0592790f0f1da7a64107959/lib/faraday/connection.rb#L456); `headers.dup` comes from [`@headers`](https://github.com/lostisland/faraday/blob/18524c6e89071fbaf0592790f0f1da7a64107959/lib/faraday/connection.rb#L72). How about changing it like this?

```suggestion
    def headers: () -> Utils::Headers
```

yykamei (commented):
Thank you! It looks great 👍 I commented a suggestion in your changes. Please take a look at it.

gems/faraday/2.5/faraday.rbs
creamstew: @yykamei
 
Thank you! I agree, so I've made the changes accordingly. Could you please review it again?
https://github.com/ruby/gem_rbs_collection/pull/628/commits/040c09e428a45e70f8973d9de9674b8527607ae5

yykamei (commented):
Nice 👍 

APPROVE!

Issue #619, opened 2024-07-08
Tests on activerecord v6.1 fail


fabon-f:
When I executed `bin/test gems/activerecord/6.1` an error occurred:

```
Testing gems/activerecord/6.1...
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Writing lockfile to ~/path/to/gem_rbs_collection/gems/activerecord/6.1/_test/Gemfile.lock
Using activemodel:6.0 (~/path/to/gem_rbs_collection/gems/activemodel/6.0)
Using activerecord:6.1 (~/path/to/gem_rbs_collection/gems/activerecord/6.1)
Using activesupport:6.0 (~/path/to/gem_rbs_collection/gems/activesupport/6.0)
Using concurrent-ruby:1.1 (~/path/to/gem_rbs_collection/gems/concurrent-ruby/1.1)
Using date:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/date/0)
Using erb:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/erb/0)
Using i18n:1.10 (~/path/to/gem_rbs_collection/gems/i18n/1.10)
Using logger:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/logger/0)
Using minitest:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/minitest/0)
Using monitor:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/monitor/0)
Using mutex_m:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/mutex_m/0)
Using pathname:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/pathname/0)
Using securerandom:0 (~/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/stdlib/securerandom/0)
[...]

PR #631, opened 2024-07-28
Fixed steep error points in ActiveRecord 6.1

rhiroe:
resolved https://github.com/ruby/gem_rbs_collection/issues/619

- Symlink of _test/metadata.yaml in 6.0 to 6.1
- Annotated `User#articles` definition using `@dynamic` and removed _test/steep_expectations.yml


rhiroe:
I am not sure how Steep errors came to be ignored.
@tk0miya Any supplemental information would be appreciated.

rhiroe:
Maybe it would be better to create a symbolic link instead of adding a file.

tk0miya:
It's much better to use symlinks for these files than copying them.

tk0miya:
Looks good. Thanks!

ksss:
APPROVE

ksss (commented):
APPROVE

PR #638, opened 2024-08-23
fix: Correct RBS for ActionController::Parameters#fetch method

7riatsu:
Updated the RBS for the `fetch` method in `ActionController::Parameters` to reflect that the block is optional.
The previous RBS incorrectly required a block, but according to the implementation, the block is optional.

Ref: https://api.rubyonrails.org/v6.0/classes/ActionController/Parameters.html#method-i-fetch


ksss (commented):
APPROVE

PR #640, opened 2024-08-24
Add the type of allow_browser in actionpack

yykamei:
Rails 7.2 adds `allow_browser` to disallow browsers that don't satisfy the version specified in the `:versions` option.

https://edgeguides.rubyonrails.org/7_2_release_notes.html#add-browser-version-guard-by-default

This method is included in the generated `ApplicationController` if a developer runs `rails new` since Rails 7.2, so I think it's worth being added in gem_rbs_collection.

This patch also includes symbolic links to RBS files located in `gems/actionpack/6.0`.


ksss (commented):
APPROVE

yykamei:
Thank you!

/merge

PR #641, opened 2024-08-26
activemodel: Remove ActiveModel::Attributes#initialize

tk0miya:
Now `ActiveModel::Attributes#initialize` has been typed as `() -> untyped`. But it's incorrect. The original implementation takes positional arguments via varargs and passes it to other constructors using `super` method. So it does not care the number of arguments.

refs:

* https://github.com/rails/rails/blob/v6.1.7.8/activemodel/lib/active_model/attributes.rb#L75
* https://github.com/rails/rails/blob/v7.2.1/activemodel/lib/active_model/attributes.rb#L106

Unfortunately, the current type conceals the correct type of the constructor.
Usually, many models using ActiveModel::Model takes keyword arguments (ex. `Person.new(name: 'John')`. But, unfortunately, the current type of ActiveModel::Atributes conceals the type of constructors for them.

This comments out the definition of ActiveModel::Attributes#initialize not to conceal the type of constructor.

### DIFF
--- a/gems/activemodel/6.0/activemodel-generated.rbs
+++ b/gems/activemodel/6.0/activemodel-generated.rbs
@@ -852,8 +852,6 @@ module ActiveModel
       def define_default_attribute: (untyped name, untyped value, untyped `type`) -> untyped
     end
 
-    def initialize: () -> untyped
-
     # Returns a hash of all the attributes with their names as keys and the values of the attributes as values.
     #
     #   class Person
--- a/gems/activemodel/6.0/_test/test.rb
+++ b/gems/activemodel/6.0/_test/test.rb
@@ -9,4 +9,5 @@ class Person
   def foo? = true
 end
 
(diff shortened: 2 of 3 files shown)


gems/activemodel/6.0/activemodel-generated.rbs
```diff
@@ -852,7 +852,7 @@ module ActiveModel
       def define_default_attribute: (untyped name, untyped value, untyped `type`) -> untyped
     end
 
-    def initialize: () -> untyped
+    # def initialize: () -> untyped
```
ksss, line 855: I think this line should be deleted.

tk0miya: done.

ksss (commented):
APPROVE

tk0miya:
Thanks!

/merge

PR #646, opened 2024-08-29
activesupport: Add ActiveSupport::TimeWithZone#to_fs

tk0miya:
refs:

* https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html#method-i-to_fs
* https://github.com/rails/rails/blob/v7.2.1/activesupport/lib/active_support/time_with_zone.rb#L212


ksss (commented):
APPROVE

PR #653, opened 2024-09-07
activerecord: Add missing methods to mixin modules

tk0miya:
`#sum` and `#scope` were only added to interfaces, but not to mixin modules.  It must be also added to the latter ones.

refs: #412, #451


ksss (commented):
APPROVE

PR #654, opened 2024-09-07
activerecord: Add activerecord-7.2

tk0miya:
As preparation for adding a new definition only for 7.2, this adds activerecord/7.2 directory.

Please let me know if my work is not good.

Thanks,


gems/activerecord/7.2/_test/test.rb
tk0miya, line 1: I copied these leading 3 files from activerecord/7.1 directory, and created symlinks for other files.
I'm not sure the copying strategy is better.

ksss (commented):
APPROVE

PR #655, opened 2024-09-07
activerecord: Arel::Table.new takes table kwargs since v6.1

tk0miya:
refs:

* https://github.com/rails/rails/pull/39881
* https://github.com/rails/rails/blob/v6.1.7.8/activerecord/lib/arel/table.rb#L17


ksss (commented):
APPROVE

tk0miya:
Thanks!

PR #657, opened 2024-09-08
activerecord: AR::Relation#new should take block argument as optional

tk0miya:
Now the block argument for `AR::Relation#new` is defined as required, but it must be optional.

refs:

* https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-new
* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/relation.rb#L69


ksss (commented):
APPROVE

PR #658, opened 2024-09-08
activerecord: Support insert_all on relation

tk0miya:
ActiveRecord supports insert and upsert methods on relation classes (AssociationRelation and CollectionProxy) since v6.1.

refs:

* https://github.com/rails/rails/pull/38899
* https://github.com/rails/rails/blob/v6.1.7.8/activerecord/CHANGELOG.md

Note:

The implementation will be moved into AR::Relation since v7.2.


ksss (commented):
APPROVE

PR #662, opened 2024-09-11
activerecord: Add ActiveRecord::Querying#with_recursive

tk0miya:
refs:

* https://github.com/rails/rails/pull/51601
* https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-with_recursive


ksss (commented):
APPROVE

PR #671, opened 2024-09-24
Add signeture of Faraday::Connection#ssl

rhiroe:
Add `Faraday::Connection#ssl` signature and `Faraday::SSLOptions` class.

After merging this Pull Request, the following code passes type checking.
```rb
conn = Faraday.new do |builder|
  builder.ssl.client_cert = 'client_cert'
  builder.ssl.client_key = 'client_key'
end
```


yykamei (commented):
Nice!

APPROVE

PR #673, opened 2024-09-24
Add signeture of Faraday::Env

rhiroe:
Add signature of `Faraday::Env` class.

After merging this Pull Request, the following code passes type checking.
```rb
Faraday::Middleware.new(Object.new, { foo: 123 }).call(Faraday::Env.new)
```


yykamei (commented):
Thank you! When I was adding Faraday signatures, there were too many untyped, and I gave up building `Env`. I appreciate your great work!

APPROVE

PR #674, opened 2024-09-24
Add ActionDispatch::Http::MimeNegotiation::InvalidType

rhiroe:
Add `ActionDispatch::Http::MimeNegotiation::InvalidType` class.

https://github.com/rails/rails/blob/d6bf4de7dc2c285d4b60d0c6da1c92668b074ae3/actionpack/lib/action_dispatch/http/mime_negotiation.rb#L12


ksss (commented):
APPROVE

PR #675, opened 2024-09-30
activerecord: Add ActiveRecord::Base.offset

tk0miya:
refs:

* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/querying.rb#L15
* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/relation/query_methods.rb#L746


ksss (commented):
APPROVE

PR #677, opened 2024-09-30
activerecord: Add #arel_table to AR::Relation

tk0miya:
The ActiveRecord::Relation provides methods defined at AR::Base via delegation (see ActiveRecord::Delegation).

Therefore AR::Relation also provides `#arel_table`.  It is often used in the scope definition.


ksss (commented):
APPROVE

PR #680, opened 2024-09-30
activerecord: Arel tree managers can take a table on initialize since v7.0

tk0miya:
refs:

* https://github.com/rails/rails/pull/41579/files
* https://github.com/rails/rails/blob/v7.0.8.4/activerecord/lib/arel/insert_manager.rb


ksss (commented):
APPROVE

tk0miya:
Thanks as always!

PR #683, opened 2024-10-03
Improve Array.wrap in ActiveSupport

nickjer:
This change better reflects the return value of `Array.wrap`.

It handles the three cases outlined in the original code:

https://github.com/rails/rails/blob/ced86070009881362446b5dbbaa36bde36c64b6a/activesupport/lib/active_support/core_ext/array/wrap.rb#L39-L47

- `Array.wrap(nil) #=> []`
- `Array.wrap([1, 2]) #=> [1, 2]`
- `Array.wrap("string") #=> ["string"]`

I used the RBS code for `Array(...)` as inspiration:

https://github.com/ruby/rbs/blob/8afce66c2a9a5e0919af9671a715e9f126481938/core/kernel.rbs#L442-L444

### DIFF
--- a/gems/activesupport/6.0/_test/test.rb
+++ b/gems/activesupport/6.0/_test/test.rb
@@ -70,3 +70,8 @@ itself if Object.new.blank? \
 "Matz".presence_in("Matzuyama")&.in?(%w[Yukihiro Matz Matsumoto])
 ActiveSupport::TimeZone['Asia/Tokyo'].to_s
 Time.find_zone(Object.name)
+
+Array.wrap(nil)
+Array.wrap([1, 2, 3])
+Array.wrap("hello")
+Array.wrap({a: 1, b: 2})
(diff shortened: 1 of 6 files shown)


ksss:
`Array.wrap` has been implemented since 6.0 and the implementation has not changed for more than 10 years.
Signatures that have not changed from version to version are appropriate to be written at https://github.com/ruby/gem_rbs_collection/blob/main/gems/activesupport/6.0/activesupport.rbs.

(Sorry for the lack of clarity.)

nickjer:
Oh, that is definitely good to know. I didn't quite know how all the files combined together in the end. I've moved it to the suggested file.

I wasn't quite sure of its location within that file. I'd be happy to move it to a better place within that file if you want.

ksss (commented):
APPROVE

nickjer:
Thanks for the review @ksss

PR #688, opened 2024-10-08
activesupport: Update the type of Array.wrap

tk0miya:
At present, `Array.wrap` will be return an empty tuple for untyped value because it matches to the first method_type: `(nil) -> []`.

In the type world, the behavior of an empty tuple is different from an empty array.  For example, calling a enumerable method will cause type an error:

```
app/app.rb:18:28: [error] Type `bot` does not have method `to_s`
│
└ Array.wrap(foo).map { |e| e.to_s }
                              ~~~~

```

This changes the return type of `Array.wrap` for nil value to `Array[untyped]`.

refs: https://github.com/ruby/gem_rbs_collection/pull/683

### DIFF
--- a/gems/activesupport/6.0/activesupport.rbs
+++ b/gems/activesupport/6.0/activesupport.rbs
@@ -501,7 +501,7 @@ class Array[unchecked out Elem]
   #
   # The differences with <tt>Kernel#Array</tt> explained above
   # apply to the rest of <tt>object</tt>s.
-  def self.wrap: (nil) -> []
+  def self.wrap: (nil) -> Array[untyped]
                | [T] (_ToAry[T] array_like) -> Array[T]
                | [T] (T ele) -> [T]
 end
--- a/gems/activesupport/6.0/activesupport.rbs
+++ b/gems/activesupport/6.0/activesupport.rbs
@@ -503,7 +503,7 @@ class Array[unchecked out Elem]
   # apply to the rest of <tt>object</tt>s.
   def self.wrap: (nil) -> Array[untyped]
                | [T] (_ToAry[T] array_like) -> Array[T]
-               | [T] (T ele) -> [T]
+               | [T] (T ele) -> Array[T]
 end
 
 module ActiveSupport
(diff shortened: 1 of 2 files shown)


tk0miya:
@nickjer What do you think about this change?

gems/activesupport/6.0/activesupport.rbs
```diff
@@ -501,7 +501,7 @@ class Array[unchecked out Elem]
   #
   # The differences with <tt>Kernel#Array</tt> explained above
   # apply to the rest of <tt>object</tt>s.
-  def self.wrap: (nil) -> []
+  def self.wrap: (nil) -> Array[untyped]
                | [T] (_ToAry[T] array_like) -> Array[T]
                | [T] (T ele) -> [T]
```
tk0miya, line 506: I have another idea to avoid the problem. It is reordering the definition not to match `untyped` to `(nil) -> []`.

```suggestion
  def self.wrap: [T] (_ToAry[T] array_like) -> Array[T]
               | (nil) -> Array[untyped]
               | [T] (T ele) -> [T]
```

This will also resolve my case.  But it will still not go well in the case of nil.

nickjer: Are you saying your original solution in this PR does not resolve your issue?

I think your original suggestion is fine although I haven't dug too deep into all of its use cases. It is unfortunate that there are no overloaded method definitions for the empty array `[]`.

I did confirm that `Array(nil).map(&:to_sym)` also blows up.

tk0miya: Oh, sorry for confusing you. My first proposal resolves my problem. So please forget this approach.

ksss, line 506: Should we also change `[T]` to `Array[T]`? 

It seems that this method is not intended to return a tuple, but rather it just happens to result in an array with one element. Given the name `Array.wrap`, I think it makes more sense for it to return an `Array`, don't you?
```suggestion
               | [T] (T ele) -> Array[T]
```

tk0miya: Reasonable. I'm okay to change it.

tk0miya: changed.

ksss (commented):
APPROVE

tk0miya:
Thank you for reviewing! @ksss @nickjer

PR #689, opened 2024-10-09
Support Symbol value for :block in #allow_browser

yykamei:
Thanks to https://github.com/rails/rails/pull/53158, the `:block` option now supports Symbol value.

This patch follows the upstream changes.


yykamei:
Maybe, the changes are correct for Rails 8.0, not Rails 7.2. I reopen this pull request later after Rails 8 is released!

PR #691, opened 2024-10-16
activerecord: Add AR::Relation#sanitize_sql_like

tk0miya:
The ActiveRecord::Relation provides methods defined at AR::Base via delegation (see ActiveRecord::Delegation).

This allows to use sanitize_sql_like and its family via AR::Relation. It is often used in the scope definition.


ksss (commented):
APPROVE

PR #693, opened 2024-10-20
railties: Add Rails::Railties.initializers

tk0miya:
We need to call "extend ClassMethods" manually to add `Rails::Railties.initializers`.

### DIFF
--- a/gems/railties/6.0/railties.rbs
+++ b/gems/railties/6.0/railties.rbs
@@ -6,3 +6,9 @@ module Rails
     end
   end
 end
+
+module Rails
+  class Railties
+    extend Initializable::ClassMethods
+  end
+end


gems/railties/6.0/patch.rbs
```diff
@@ -8,6 +8,12 @@ module Rails
   end
 end
 
+module Rails
+  class Railties
+    extend Initializable::ClassMethods
+  end
+end
+
```
ksss, line 16: I think, We should create **railtie.rbs**.
It is a file written rbs, It's not generated and not patch.
How about you?

tk0miya: +1 for creating railtie.rbs.

> It is a file written rbs, It's not generated and not patch.

It would be better to discuss what is a "patch".
Because we have many patch.rbs. 

```
$ tkomiya@tarf> find gems -name patch.rbs
gems/actionpack/7.2/patch.rbs
gems/actionpack/6.0/patch.rbs
gems/activesupport/6.0/patch.rbs
gems/activesupport/7.0/patch.rbs
gems/activemodel/6.0/patch.rbs
gems/activemodel/7.0/patch.rbs
gems/activejob/6.0/patch.rbs
gems/railties/6.0/patch.rbs
gems/activerecord/7.2/patch.rbs
gems/activerecord/6.1/patch.rbs
gems/activerecord/6.0/patch.rbs
gems/activerecord/7.1/patch.rbs
gems/activerecord/7.0/patch.rbs
gems/nokogiri/1.11/patch.rbs
gems/actionview/6.0/patch.rbs
gems/enumerize/2.5/patch.rbs
```

I think definitions for the gem itself should be moved to [gem_name].rbs or something else.
And patches for other gems should be kept in patch.rbs.

In railties gem's case, ERB, Minitest, Rack and so on keep be placed on here, right?

tk0miya: As a first step, I moved the definition of `Rails::Generator::PluginGenerator#app_path` to railties.rbs
https://github.com/ruby/gem_rbs_collection/pull/697

ksss: I think of `patch.rbs` as being like `.rubocop_todo.yml`. We should try to reduce it as much as possible.

tk0miya:
Now I moved the new definition to railities.rbs

ksss (commented):
APPROVE

PR #695, opened 2024-10-22
activemodel: Add ActiveModel::API

tk0miya:
ActiveModel::API has been added since v7.0.

refs: https://github.com/rails/rails/pull/43223


ksss (commented):
APPROVE

PR #700, opened 2024-10-27
activerecord: Allow to use array of Hashes to #preload

tk0miya:
The real use case is here: https://github.com/kaigionrails/conference-app/pull/341/files#r1817211252

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -355,9 +355,9 @@ module ActiveRecord
       def joins: (*String | Symbol | Hash[untyped, untyped]) -> self
       def left_joins: (*String | Symbol | Hash[untyped, untyped]) -> self
       def left_outer_joins: (*String | Symbol | Hash[untyped, untyped]) -> self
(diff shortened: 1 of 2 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -296,6 +296,8 @@ module ActiveRecord
   end
 
   module QueryMethods
+    type includee = String | Symbol | Array[includee] | Hash[untyped, untyped]
```
tk0miya, line 299: I'm not sure this is the best place for this type alias. I'd really appreciate if you give me your opinion.
Thanks,

tk0miya: I don't have a strong opinion of accuracy. So I'm okay to stop using recurrent definitions like this.

```suggestion
    type includee = String | Symbol | Array[untyped] | Hash[untyped, untyped]
```

tk0miya: Finally, I removed `includee` type alias. It's only needed for recurrent definitions.

ksss (commented):
APPROVE

PR #701, opened 2024-10-28
activestorage: Add types for 7.0

tk0miya:
As preparation for adding a new class, this adds gems/activestorage/7.0 directory.

Which is better adding 7.0 or 7.2? Either is okay with me.


ksss:
> Which is better adding 7.0 or 7.2? Either is okay with me.
It depends on what you want to add.

APPROVE

ksss (commented):
APPROVE

tk0miya:
My goal is https://github.com/ruby/gem_rbs_collection/pull/702.
I thought it's better to provide types for wide people. So I'm going with this.

PR #702, opened 2024-10-28
activestorage: Add ActiveStorage::Current

tk0miya:
This contains https://github.com/ruby/gem_rbs_collection/pull/701.

refs:
    
* https://github.com/rails/rails/blob/v6.0.6.1/activestorage/app/models/active_storage/current.rb
* https://github.com/rails/rails/blob/v7.0.8.6/activestorage/app/models/active_storage/current.rb
    * `#url_options` has been added and `#host` has been deprecated


ksss (commented):
APPROVE

tk0miya:
Thanks!

PR #706, opened 2024-10-28
actionpack: Remove unnecessary patches to other gems

tk0miya:


gems/actionpack/6.0/patch.rbs
```diff
@@ -47,13 +47,6 @@ module Capybara
   end
 end
 
-# Remove the fake types for Minitest
-# if the real types are available.
-module Minitest
-  module Assertions
-  end
-end
```
tk0miya, line 55: https://github.com/ruby/rbs/blob/master/stdlib/minitest/0/minitest/assertions.rbs

gems/actionpack/6.0/patch.rbs
```diff
@@ -82,20 +75,3 @@ module Rack
     end
   end
 end
-
-# Remove the fake types for Gem::Version
-# if the real types are available.
-module Gem
-  class Version
-  end
-end
```
tk0miya, line 91: https://github.com/ruby/rbs/blob/master/core/rubygems/version.rbs

gems/actionpack/6.0/patch.rbs
```diff
@@ -82,20 +75,3 @@ module Rack
...
-
-# Remove the fake types for SimpleDelegator
-# if the real types are available.
-class SimpleDelegator
-end
```
tk0miya, line 96: https://github.com/ruby/rbs/blob/master/stdlib/delegate/0/simple_delegator.rbs

gems/actionpack/6.0/patch.rbs
```diff
@@ -82,20 +75,3 @@ module Rack
...
-
-module Rails
-  class Engine
-  end
-end
```
tk0miya, line 101: https://github.com/ruby/gem_rbs_collection/blob/main/gems/railties/6.0/railties-generated.rbs#L1947

Note: It seems not used. So I only removed this.

ksss (commented):
APPROVE

PR #710, opened 2024-10-30
activemodel: Remove unnecessary patches to other gems

tk0miya:


gems/activemodel/6.0/patch.rbs
```diff
@@ -1,6 +0,0 @@
-# Remove the fake types for Gem::Version
-# if the real types are available.
-module Gem
-  class Version
-  end
-end
```
tk0miya, line 6: https://github.com/ruby/rbs/blob/master/core/rubygems/version.rbs#L163

ksss (commented):
APPROVE

PR #711, opened 2024-10-30
activerecord: Remove unnecessary patches to other gems

tk0miya:


gems/activerecord/6.0/patch.rbs
```diff
@@ -27,15 +27,3 @@ module PG
...
-# if the real types are available.
-class SimpleDelegator
-end
-
-# Remove the fake types for Gem::Version
-# if the real types are available.
-module Gem
-  class Version
-  end
-end
```
tk0miya, line 41: https://github.com/ruby/rbs/blob/master/core/rubygems/version.rbs#L163

gems/activerecord/6.0/patch.rbs
```diff
@@ -27,15 +27,3 @@ module PG
     def async_exec: (*untyped) -> untyped
   end
 end
-
-# Remove the fake SimpleDelegator type
-# if the real types are available.
-class SimpleDelegator
-end
```
tk0miya, line 34: https://github.com/ruby/rbs/blob/master/stdlib/delegate/0/simple_delegator.rbs#L69

gems/activerecord/6.1/manifest.yaml
tk0miya, line 1: Is it better to copy the manifest.yaml? 7.* directories have their copy (not symlink).

ksss: Perhaps a copy is better.

ksss (commented):
APPROVE

tk0miya:
Now I replaced the symlink with a real file.

ksss:
LGTM!

tk0miya:
@ksss re-APPROVE is needed after the change. Could you review this again?
Thanks as always.

ksss (commented):
APPROVE

PR #713, opened 2024-10-31
Add type definitions to delete_all

kazuyainoue0124:
Fix types definitions for `delete_all` methods.
Changed the return type of `delete_all` methods to `Integer`.

### DIFF
--- a/gems/activerecord/6.0/activerecord-generated.rbs
+++ b/gems/activerecord/6.0/activerecord-generated.rbs
@@ -896,7 +896,7 @@ module ActiveRecord
       # @author.books.delete_all(:delete_all)
       #
       # See delete for more info.
-      def delete_all: (?untyped? dependent) -> untyped
+      def delete_all: (?Symbol dependent) -> Integer
 
       # Destroy all the records from this association.
       #
@@ -1232,7 +1232,7 @@ module ActiveRecord
       #
       #   Pet.find(1, 2, 3)
       #   # => ActiveRecord::RecordNotFound: Couldn't find all Pets with 'id': (1, 2, 3)
-      def delete_all: (?untyped? dependent) -> untyped
+      def delete_all: (?Symbol dependent) -> Integer
 
       # Deletes the records of the collection directly from the database
       # ignoring the +:dependent+ option. Records are instantiated and it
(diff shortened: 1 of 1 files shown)


tk0miya:
FYI: API doc mentions the return value of `#delete_all`.

> Returns the number of rows affected.
> https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all

ksss (commented):
How was the effect of this change?

Perhaps gems/activerecord/6.0/activerecord.rbs also needs to be modified for substantial effect.

tk0miya:
I noticed `ActiveRecord::Associations::CollectionProxy#delete_all` takes an optional argument `dependent`. But `ActiveRecord::Associations::Relation#delete_all` does not take arguments at all.

But the current design of rbs_rails expects both to be the same.
Therefore it's not good to modify definitions under activerecord.rbs (`ActiveRecord::Relation::Methods`, `ActiveRecord::Base::ClassMethods`, `_ActiveRecord_Relation` and `_ActiveRecord_Relation_ClassMethods`).

It's better to change only the return types of these methods under activerecord.rbs.
And we need to hack rbs_rails to override the argument type of CollectionProxy#delete_all via rbs_rails like https://github.com/pocke/rbs_rails/pull/289 does.


Refs:

* https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete_all
* https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all

kazuyainoue0124:
Thank you for your valuable feedback.
I apologize for not fully understanding the differences between `ActiveRecord::Associations::CollectionProxy#delete_all` and `ActiveRecord::Associations::Relation#delete_all`, as well as the specifics of `rbs_rails`.

In this pull request, I have only updated the return types.
I will address the override of `CollectionProxy#delete_all` in `rbs_rails` separately.

If there are any remaining issues, please let me know.

ksss:
Sorry for the delay in responding.

- When modifying 6.0/activerecord-generated.rbs, move as far as possible to 6.0/activerecord.rbs.
- I would like at least one line of testing.

PR #715, opened 2024-11-06
actionpack: Fix helper and fragment_cache_key block to optional

sanfrecce-osaka:


gems/actionpack/6.0/actioncontroller.rbs
```diff
@@ -25,6 +25,8 @@ module ActionController
     include EtagWithTemplateDigest
     include EtagWithFlash
     include Caching
+    extend AbstractController::Caching::ClassMethods
+    extend AbstractController::Caching::Fragments::ClassMethods
```
tk0miya, line 29: It would be better to not only extend ClassMethods but also include `AbstractController::Caching` here.

`ActionController::Caching` dynamically calls include inside the included block. Therefore it's better to include it explicitly.

https://github.com/rails/rails/blob/v6.0.6.1/actionpack/lib/action_controller/base.rb#L219
https://github.com/rails/rails/blob/v6.0.6.1/actionpack/lib/action_controller/caching.rb#L28-L30
https://github.com/rails/rails/blob/v6.0.6.1/actionpack/lib/abstract_controller/caching.rb#L46

gems/actionpack/6.0/actioncontroller.rbs
```diff
@@ -172,3 +174,275 @@ module ActionController
...
+      #       "#{bar} is the very best"
+      #     end
+      #   end
+      #
+      # Finally, all the above styles can be mixed together, and the +helper+ method can be invoked with a mix of
+      # +symbols+, +strings+, +modules+ and blocks.
+      #
+      #   helper(:three, BlindHelper) { def mice() 'mice' end }
+      #
+      def helper: (*untyped args) ?{ () -> untyped } -> untyped
```
tk0miya, line 249: Could you move the methods only what you changed, please? If my understanding is correct, only `ActionController::Helpers::ClassMethods#helper` and `AbstractController::Caching::Fragments::ClassMethods#fragment_cache_key` should be moved and modified.

gems/actionpack/6.0/actioncontroller.rbs
sanfrecce-osaka: Thank you for your feedback, I fixed it in [6b169ec](https://github.com/ruby/gem_rbs_collection/pull/715/commits/6b169ec2f39abf9ee212ac721c9e2f2e6e416e13)

gems/actionpack/6.0/actioncontroller.rbs
sanfrecce-osaka: Sorry 🙏 
I wasn't sure how far I should move the auto-generated code, including comments, so I made https://github.com/sanfrecce-osaka/gem_rbs_collection/commit/d7a4fb518abd0c258ea37b46d796bdcc31ba7e1a and https://github.com/sanfrecce-osaka/gem_rbs_collection/commit/08b92b9e5ab942637942e3736972492ad9f5bff0.
I fixed in [939ac11](https://github.com/ruby/gem_rbs_collection/pull/715/commits/939ac1144fa8335cb2233c9b434da784ec2c259f) based on your feedback.

gems/actionpack/6.0/actioncontroller.rbs
```diff
@@ -172,3 +175,83 @@ module ActionController
...
```
tk0miya, line 216: Note: The above comment describes the argument are Module, Symbol, String. So it would be better to update signature like this:

```suggestion
      def helper: (*Module | Symbol | String args) ?{ () -> void } -> void
```


Either way, this is off-topic. I'll post another PR after merging this.

tk0miya:
APPROVE

tk0miya:
APPROVE

tk0miya (commented):
APPROVE

tk0miya:
Sorry for the confusion. I did not know the proper way to send APPROVE.

PR #717, opened 2024-11-10
activemodel: ActiveModel::Errors#add accepts String and Proc for type

sanfrecce-osaka:
The behavior of passing String or Proc for type should be allowed.

FYI: It is documented in 6.1, but It is already available in 6.0.

cf. https://github.com/rails/rails/commit/fcd1e41e82d3c9993f96e1763ac49a196678f931#diff-4deafad6fefcf6aa6626a1f7e3cd0ca1cf95e358f55a10a69e02be84d926bb31R368-R374

cf. https://github.com/rails/rails/commit/d9011e39357300fe78720227af4c13b4bc4ac4dd#diff-4deafad6fefcf6aa6626a1f7e3cd0ca1cf95e358f55a10a69e02be84d926bb31R370

Also, the name of the second argument was changed from `message` to `type` in 6.1.


gems/activemodel/6.0/activemodel.rbs
```diff
@@ -498,3 +498,50 @@ module ActiveModel
...
+    # +attribute+ should be set to <tt>:base</tt> if the error is not
+    # directly associated with a single attribute.
+    #
+    #   person.errors.add(:base, :name_or_email_blank,
+    #     message: "either name or email must be present")
+    #   person.errors.messages
+    #   # => {:base=>["either name or email must be present"]}
+    #   person.errors.details
+    #   # => {:base=>[{error: :name_or_email_blank}]}
+    def add: (untyped attribute, ?::Symbol | ::String | ^(untyped base, ::Hash[untyped, untyped]) -> (::Symbol | ::String) type, ?::Hash[untyped, untyped] options) -> void
```
ksss, line 545: What makes you think return type should be `void` instead of `ActiveModel::Error`?

sanfrecce-osaka: Thank you for your feedback.
ActiveModel::Error was added on 6.1.

cf. https://github.com/rails/rails/commit/ef68d3e35cb58f9f491993eeec6e7de99442dd06

On 6.1 or later, ActiveModel::Errors#add returns ActiveModel::Error.
On the other hand, `ActiveModel::Errors#add` returns messages array on rails 6.0, because last evaluated is `Array#<<`.

cf. https://github.com/rails/rails/blob/v6.0.0/activemodel/lib/active_model/errors.rb#L311-L322

Currently in gem_rbs_collection, activemodel type definitions are defined 6.0 and 7.0, but not 6.1.

cf. https://github.com/ruby/gem_rbs_collection/pull/615

So I fixed them on [4981ed6](https://github.com/ruby/gem_rbs_collection/pull/717/commits/4981ed698b8b26a7d21d40c2eb811501bee7c552)

ksss (approved):
APPROVE

sanfrecce-osaka:
@ksss 

Sorry, I resolved conflict. Please re-review 🙏 🙏 🙏

ksss (commented):
APPROVE

PR #718, opened 2024-11-10
activerecord: CollectionProxy#build can take an array of EachPair

tk0miya:
As described above, `CollectionProxy#build` takes an array of attributes and returns list of instances of the model.

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -826,6 +826,7 @@ module ActiveRecord
       #   person.pets.size  # => 5 # size of the collection
       #   person.pets.count # => 0 # count from database
       def build: (?_EachPair attributes) ?{ () -> untyped } -> untyped
+               | (Array[_EachPair] attributes) ?{ () -> untyped } -> Array[untyped]
 
       alias new build
 


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -825,7 +825,8 @@ module ActiveRecord
       #
       #   person.pets.size  # => 5 # size of the collection
       #   person.pets.count # => 0 # count from database
-      def build: (?_EachPair attributes) ?{ () -> untyped } -> untyped
+      def build: (?Array[_EachPair] attributes) ?{ () -> untyped } -> Array[untyped]
+               | (?_EachPair attributes) ?{ () -> untyped } -> untyped
```
ksss, line 829: It looks like you have created the possibility to return an Array even when there are no arguments, is this intentional?

```rb
person.pets.build #=> Array[untyped]
```

tk0miya: Good point. It's my mistake. Fixed now.

tk0miya: Finally I appended the Array case to the tail. It's better to return a single entry if `untyped` value given.

ksss (commented):
APPROVE

tk0miya:
Thanks!

PR #719, opened 2024-11-10
activerecord: Add ActiveRecord::Base.count

tk0miya:
refs:

* https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count
* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/relation/calculations.rb#L41
* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/querying.rb#L18


ksss (commented):
APPROVE

PR #720, opened 2024-11-10
activerecord: Fix the location of #with_recursive

tk0miya:
It is defined at `ActiveRecord::QueryMethods`, not `ActiveRecord:: Relation::QueryMethods`.

Unfortunately, it works well even if the wrong location.  But it conceals other methods defined in `AR::QueryMethods`.

refs: #662

### DIFF
--- a/gems/activerecord/7.2/activerecord-7.2.rbs
+++ b/gems/activerecord/7.2/activerecord-7.2.rbs
@@ -19,13 +19,6 @@ module ActiveRecord
     def upsert_all: (untyped attributes, ?unique_by: untyped?, ?returning: untyped?, ?record_timestamps: bool?) -> untyped
   end
 
-  class Relation
-    module QueryMethods
-      def with_recursive: (*untyped) -> untyped
-      def with_recursive!: (*untyped) -> untyped
-    end
-  end
-
   module Inheritance
     module ClassMethods
       def primary_abstract_class: () -> void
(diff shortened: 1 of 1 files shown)


gems/activerecord/7.2/activerecord-7.2.rbs
```diff
@@ -46,6 +39,11 @@ module ActiveRecord
     extend RuntimeRegistry
   end
 
+  module QueryMethods
+    def with_recursive: (*untyped) -> untyped
+    def with_recursive!: (*untyped) -> untyped
+  end
+
   module QueryMethods
```
ksss, line 47: How about merging `module QueryMethods`?

tk0miya: I'm worried about the sub-modules and classes becoming bigger.

```
module QueryMethods
  class WhereChain[Relation]
     (long long definitions)
   end

  module OtherMod
    (also very long)
  end

  def with_recursive: (*untyped) -> untyped
  def with_recursive!: (*untyped) -> untyped
end
```

I know I worry too much, but activerecord is a large library.
Therefore, I think it's better to separate definitions per module or class.
For example, `ActiveRecord::Relation` and `ActiveRecord::Associations` have many sub-modules.

ksss (approved):
APPROVE

PR #659, opened 2024-09-08
activerecord: Add ActiveRecord::Base.generates_tokens_for

tk0miya:
`ActiveRecord::Base.generates_token_for` has been added since v7.1.

refs:

* https://github.com/rails/rails/pull/44189
* https://github.com/rails/rails/blob/v7.1.4/activerecord/CHANGELOG.md

PR #721, opened 2024-11-10
activerecord: Add ActiveRecord::TokenFor::RelationMethods

tk0miya:
refs:

* #659
* https://github.com/rails/rails/pull/51801
* https://api.rubyonrails.org/classes/ActiveRecord/TokenFor/RelationMethods.html
* https://github.com/rails/rails/blob/v7.2.2/activerecord/lib/active_record/token_for.rb


ksss (commented):
APPROVE

PR #723, opened 2024-11-16
activemodel: Add gems/activemodel/7.1

tk0miya:
Create symlinks and copies from 6.0 and 7.0. Nothing changed yet.

PR #724, opened 2024-11-16
activemodel: Add types for ActiveModel::Access

tk0miya:
It was added since v7.1

refs: https://github.com/rails/rails/pull/44544

Note: This contains https://github.com/ruby/gem_rbs_collection/pull/723.


ksss (commented):
APPROVE

PR #728, opened 2024-11-17
actionview: Add types for content_for, form_with and render

tk0miya:
FYI: This PR contains two commits. The first one is moving definitions to actionview.rbs. And another one is modifying definitions to the better ones.


ksss (commented):
APPROVE

PR #730, opened 2024-11-21
add ActiveRecord::Calculations module to ActiveRecord::Base

tasuku-lab280:
I wanted to use `ActiveRecord::Calculations` methods like `User.count` as class methods of the model, so added code to extend `ActiveRecord::Calculations` module in `ActiveRecord::Base`.

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -17,6 +17,7 @@ module ActiveRecord
     extend ::ActiveRecord::AttributeMethods::Write::ClassMethods
     extend ::ActiveRecord::Attributes::ClassMethods
     extend ::ActiveRecord::AutosaveAssociation::ClassMethods
+    extend ::ActiveRecord::Calculations
     extend ::ActiveRecord::Core::ClassMethods
     extend ::ActiveRecord::CounterCache::ClassMethods
     extend ::ActiveRecord::Inheritance::ClassMethods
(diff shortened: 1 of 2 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -17,6 +17,7 @@ module ActiveRecord
     extend ::ActiveRecord::AttributeMethods::Write::ClassMethods
     extend ::ActiveRecord::Attributes::ClassMethods
     extend ::ActiveRecord::AutosaveAssociation::ClassMethods
+    extend ::ActiveRecord::Calculations
```
tasuku-lab280, line 20: [ref]
* https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation/calculations.rb#L94-L104
* https://github.com/ruby/gem_rbs_collection/blob/main/gems/activerecord/6.0/activerecord-generated.rbs#L17575-L17576

tk0miya:
Thank you for your contribution.

Unfortunately, this is not correct. ActiveRecord::Base does not extend `ActiveRecord::Calculations` directly.
The reason why ActiveRecord::Base provides the `.count` method is `ActiveRecord::Querying`.
It delegates tons of methods to the `.all` method.

You can see the `.count` method in the list:
https://github.com/rails/rails/blob/v8.0.0/activerecord/lib/active_record/querying.rb#L19

Therefore, extending the Calculation module in this class is not good.

Instead, in activerecord.rbs, we list such methods up to `ActiveRecord::Base::ClassMethods` and `_ActiveRecord_Relation_ClassMethods`.
https://github.com/ruby/gem_rbs_collection/blob/443b2934235be3a76b9022b21764d06429ac6d7e/gems/activerecord/6.0/activerecord.rbs#L422
https://github.com/ruby/gem_rbs_collection/blob/443b2934235be3a76b9022b21764d06429ac6d7e/gems/activerecord/6.0/activerecord.rbs#L590

tasuku-lab280:
@tk0miya 
Thanks for the review! I understand!
I will close this PR once I have implemented the type definitions based on the information you gave me, and I will remake the PR again once I have implemented the type definitions.

PR #731, opened 2024-11-23
activerecord: Support QueryObject Pattern for AR::Base.scope

tk0miya:
ActiveRecord::Base.scope can take an object having `call` method as 2nd argument "body".  QueryObject pattern is a technique to implement complicated query using a class having `call` method.

This allows to pass a QueryObject (_Callable) to the scope definition.

refs:

* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/scoping/named.rb#L171
* https://github.com/Selleo/pattern/blob/master/lib/patterns/query.rb
* https://takaokouji.github.io/output/query-object/ (Japanese)

### DIFF
--- a/gems/activerecord/6.0/_test/activerecord-generated.rb
+++ b/gems/activerecord/6.0/_test/activerecord-generated.rb
@@ -1,3 +1,8 @@
+class FunQuery
+  def self.call(*args)
+  end
+end
+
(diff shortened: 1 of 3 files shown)


ksss (commented):
APPROVE

PR #732, opened 2024-11-25
activejob: Add types for ActiveJob::Logging 

sanfrecce-osaka:
- `logger` is defined using `Class#cattr_accessor` in activesupport with instance_reader/instance_writer/instance_accessor set to true.
  - The default for `logger` is ActiveSupport::TaggedLogging, but it is not necessarily an instance of ActiveSupport::TaggedLogging or an inheritor of ActiveSupport::TaggedLogging (e.g. lograge). Therefore, it is defined as untyped.
- `log_arguments` is defined using `Class#class_attribute` in activesupport with instance_accessor set to false and instance_predicate set to true.

### DIFF
--- a/gems/activejob/6.0/activejob.rbs
+++ b/gems/activejob/6.0/activejob.rbs
@@ -8,6 +8,8 @@ module ActiveJob
     extend ActiveJob::QueueAdapter::ClassMethods
     extend ActiveJob::QueueName::ClassMethods
     extend ActiveJob::QueuePriority::ClassMethods
+    extend ActiveJob::Logging::ClassMethods
+    include ActiveJob::Logging
 
     def self.queue_as: (*untyped) -> void | ...
     def self.queue_with_priority: (?untyped? priority) -> void | ...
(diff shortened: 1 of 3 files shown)


gems/activejob/6.0/activejob.rbs
```diff
@@ -21,3 +23,27 @@ module ActiveJob
...
 end
+
+module ActiveJob
+  module Logging
+    # nodoc:
+    extend ActiveSupport::Concern
+
+    module ClassMethods
+      def logger: () -> untyped
+      def logger=: (untyped val) -> untyped
```
tk0miya, line 34: FYI: The API doc says:

> Accepts a logger conforming to the interface of Log4r or the default Ruby Logger class. You can retrieve this logger by calling logger on either an Active Job job class or an Active Job job instance.
https://api.rubyonrails.org/classes/ActiveJob/Logging.html#method-i-logger

I'm not sure "the interface of Log4r". So I'm okay to keep this untyped.

PR #734, opened 2024-11-26
activemodel: Add types for ActiveModel::AttributeRegistration

tk0miya:
It was added at v7.1

ref: https://github.com/rails/rails/pull/44664


ksss (commented):
APPROVE

PR #735, opened 2024-12-05
Various type signature updates for Rails

hjwylde:
Hi! I have been working on improving the warnings in RubyMine based on type signatures to work better, and there were a number of false warnings I found. The majority of the warnings are related to block specifications being too rigid when they should be marked as optional. I've improved the type signatures where possible, but the majority are just tiny changes to ensure that there are no false positive warnings from these signatures.

Any signatures touched were additionally moved from the generated file to the manual file (as I saw that done in other pull requests).


hjwylde:
> @hjwylde Thank you for contributing. But sorry, there are too many changes, and I feel it's difficult for me to review them.

Would it be enough to split this into different pull requests per gem?

hjwylde:
I will close this review and create a separate one for each gem.

PR #736, opened 2024-12-06
actioncontroller: signature updates

hjwylde:
Split from #735 to be smaller. Type signature updates for invalid signatures in actioncontroller.


tk0miya (commented):
APPROVE

Note:

`#key?`, `#has_key?`, `#has_value?` and `#include?` are delegated to the `@parameters`.

* https://github.com/rails/rails/blob/v6.0.6.1/actionpack/lib/action_controller/metal/strong_parameters.rb#L214-L215
* https://github.com/rails/rails/blob/v6.0.6.1/actionpack/lib/action_controller/metal/strong_parameters.rb#L243

PR #740, opened 2024-12-06
activerecord: signature updates

hjwylde:
Split from #735 to be smaller. Type signature updates for invalid signatures in activerecord, the majority are due to blocks being marked as required, but they should be optional. Another update for `NoTouching#touch` to match its declaration (see [    def touch(*, **) # :nodoc:](https://github.com/rails/rails/blob/dd8f7185faeca6ee968a6e9367f6d8601a83b8db/activerecord/lib/active_record/no_touching.rb#L61)).

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -954,3 +954,9 @@ module ActiveRecord
     end
   end
 end
+
+module ActiveRecord
+  module NoTouching
+    def touch: (*untyped, **untyped) -> untyped
+  end
+end
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -402,6 +402,7 @@ module ActiveRecord
       def offset: (Integer) -> self
       def offset!: (Integer) -> self
       def find_each: (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) { (Model) -> void } -> nil
+                   | (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) -> Enumerator[Model, nil]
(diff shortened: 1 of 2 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -513,6 +515,70 @@ module ActiveRecord
...
+    #   end
+    #
+    # NOTE: It's not possible to set the order. That is automatically set to
+    # ascending on the primary key ("id ASC") to make the batch ordering
+    # work. This also means that this method only works when the primary key is
+    # orderable (e.g. an integer or string).
+    #
+    # NOTE: By its nature, batch processing is subject to race conditions if
+    # other processes are modifying the database.
+    def find_each: (?error_on_ignore: untyped? error_on_ignore, ?batch_size: ::Integer batch_size, ?finish: untyped? finish, ?start: untyped? start) ?{ (untyped) -> untyped } -> untyped
```
tk0miya, line 578: Is this change needed for you? If you're using the rbs_rails gem, the changes of `_ActiveRecord_Relation` and `_ActiveRecord_Relation_ClassMethods` will resolve your case.

IMO, it would be better to revert this change to keep activerecord.rbs simple. What do you think?

hjwylde: We've implemented some functionality in RubyMine where users can opt in to having additional type support from `gem_rbs_collection`, without the need to use RBS in their project, see https://www.jetbrains.com/help/ruby/rbs.html#rbs_collection. It benefits users who want an improved UX in their tooling, but don't want to explicitly use RBS in their project, or perhaps cannot use RBS in their project for some reason.

For this use-case, `rbs_rails` isn't used, so there are no additional type overrides from the interfaces; I think it would be helpful to us if the default types were accurate.

tk0miya: Understand. Okay, let's go with this.

tk0miya (commented):
APPROVE

PR #742, opened 2024-12-06
railties: signature updates

hjwylde:
Split from #735 to be smaller. Type signature updates for invalid signatures in railties, due to blocks being marked as required, but they should be optional.

### DIFF
--- a/gems/railties/6.0/railties-generated.rbs
+++ b/gems/railties/6.0/railties-generated.rbs
@@ -1997,10 +1997,6 @@ module Rails
     # Defines additional Rack env configuration that is added on each call.
     def env_config: () -> untyped
 
-    # Defines the routes for this engine. If a block is given to
-    # routes, it is appended to the engine.
-    def routes: () { () -> untyped } -> untyped
-
     # Define the configuration object for the engine.
     def config: () -> untyped
 
--- a/gems/railties/6.0/railties-generated.rbs
+++ b/gems/railties/6.0/railties-generated.rbs
@@ -1434,12 +1434,6 @@ module Rails
     class MiddlewareStackProxy
       def initialize: (?untyped operations, ?untyped delete_operations) -> untyped
 
-      def insert_before: (*untyped args) { () -> untyped } -> untyped
-
-      alias insert insert_before
-
-      def insert_after: (*untyped args) { () -> untyped } -> untyped
-
       def swap: (*untyped args) { () -> untyped } -> untyped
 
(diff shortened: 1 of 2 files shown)


gems/railties/6.0/railties-generated.rbs
```diff
@@ -1434,16 +1434,8 @@ module Rails
...
-
-      def insert_after: (*untyped args) { () -> untyped } -> untyped
-
       def swap: (*untyped args) { () -> untyped } -> untyped
 
-      def use: (*untyped args) { () -> untyped } -> untyped
-
       def delete: (*untyped args) { () -> untyped } -> untyped
 
       def unshift: (*untyped args) { () -> untyped } -> untyped
```
tk0miya, line 1441: Could you change `#swap`, `#delete`, and `#unshift` too?


Note:

The args and block are forwarded to the rack middleware class when building its instance. Therefore it's better to make them optional.

refs:

* https://github.com/rails/rails/blob/v6.0.6.1/railties/lib/rails/configuration.rb#L37
* https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/stack.rb#L43-L45

gems/railties/6.0/railties.rbs
```diff
@@ -1,3 +1,25 @@
...
+      def use: (*untyped args) ?{ () -> untyped } -> untyped
+    end
+  end
+end
+
+module Rails
+  class Engine
+    # Defines the routes for this engine. If a block is given to
+    # routes, it is appended to the engine.
+    def routes: () ?{ () -> untyped } -> untyped
```
tk0miya, line 19: LGTM.

As mentioned in the above comment, the block is optional.
https://github.com/rails/rails/blob/v6.0.6.1/railties/lib/rails/engine.rb#L539

gems/railties/6.0/railties-generated.rbs
hjwylde: Thanks! I'll move those and make their arguments/block optional as well.

gems/railties/6.0/railties.rbs
hjwylde: Thanks! I believe the syntax I have used is marking the block as optional, unless you had a different syntax in mind or I've mistaken something?

tk0miya: Sorry for the confusion. The second paragraph is a note for me.

tk0miya (commented):
APPROVE

PR #746, opened 2024-12-11
Add a string type to the `lock` method arguments

sinsoku:
The `lock` method accepts a string value as well as a boolean value.

```ruby
Account.lock("FOR UPDATE NOWAIT").find_by(name: "shugo")
```

This behavior remains the same in v6.0.x.

refs:

* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/locking/pessimistic.rb
* https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/relation/query_methods.rb#L757-L770


Little-Rubyist (commented):
APPROVE

PR #749, opened 2024-12-14
actionview: Fix typo in module name: CaptureHelper

tk0miya:
Fix typo in the module name: s/CaptionHelper/CaptureHelper/.

This typo was added in https://github.com/ruby/gem_rbs_collection/pull/728.


ksss (commented):
APPROVE

PR #750, opened 2024-12-14
actionview: Add definitions for dynamic generated methods of FormBuilder

tk0miya:
refs: https://github.com/rails/rails/blob/v6.0.6.1/actionview/lib/action_view/helpers/form_helper.rb#L1904-L1914


ksss (commented):
APPROVE

PR #752, opened 2024-12-15
activerecord: add ActiveRecord::Core#values_at

dak2:
refs.
 - https://github.com/rails/rails/pull/36481
 - https://api.rubyonrails.org/v6.1.7.8/classes/ActiveRecord/Core.html#method-i-values_at
 - https://github.com/rails/rails/blob/6-1-stable/activerecord/CHANGELOG.md#rails-610-december-09-2020

### DIFF
--- a/gems/activerecord/6.1/activerecord-6.1.rbs
+++ b/gems/activerecord/6.1/activerecord-6.1.rbs
@@ -6,6 +6,10 @@ module ActiveRecord
     extend DelegatedType
   end
 
+  module Core
+    def values_at: (*Symbol | String) -> Array[untyped]
+  end
+
   class AssociationRelation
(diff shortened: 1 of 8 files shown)


gems/activerecord/6.1/activerecord-6.1.rbs
```diff
@@ -6,6 +6,10 @@ module ActiveRecord
     extend DelegatedType
   end
 
+  module Core
+    def values_at: (*Symbol | String) -> Array[untyped]
+  end
```
tk0miya, line 11: Could you make copies of this signature to other versions?
It would be nice if we could provide this type for Rails-7.0, 7.1 and 7.2.

dak2: @tk0miya 

I copied this signature to other versions without 6.0.
Thanks!

tk0miya (commented):
APPROVE

PR #760, opened 2024-12-26
activerecord: Add types for create_with to singleton that inherited activerecord base

dak2:
Singleton that inherited activerecord base can handle create_with method.

https://github.com/rails/rails/blob/6df15e25c875e9c1ee39acb0bc161b6c47531331/activerecord/test/cases/scoping/relation_scoping_test.rb#L499-L507

- create_with
  - https://api.rubyonrails.org/v6.0.6.1/classes/ActiveRecord/QueryMethods.html#method-i-create_with

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -515,6 +515,7 @@ module ActiveRecord
       def count: () -> ::Integer
               | () { (Model) -> boolish } -> ::Integer
               | (Symbol | String column_name) -> ::Integer
+      def create_with: (nil | Hash[untyped, untyped]) -> Relation
     end
   end
 end
@@ -755,6 +756,7 @@ interface _ActiveRecord_Relation_ClassMethods[Model, Relation, PrimaryKey]
   def count: () -> ::Integer
           | () { (Model) -> boolish } -> ::Integer
           | (Symbol | String column_name) -> ::Integer
+  def create_with: (nil | Hash[untyped, untyped]) -> Relation
 end
 
 module Arel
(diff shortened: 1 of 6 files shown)


gems/activerecord/6.0/activerecord-6.0.rbs
```diff
@@ -1,4 +1,10 @@
 module ActiveRecord
+  class Base
+    module ClassMethods[Model, Relation, PrimaryKey]
+      def create_with: (nil | Hash[untyped, untyped]) -> Relation
```
tk0miya, line 4: Why do you add this to each version? It seems all definitions are the same. I feel it's enough to add this to `activerecord.rbs`.

dak2: Thanks for your comment!

I didn't know that if you write in activerecord.rbs, the rbs is shared in each version.
I rewrote create_with rbs definition in activerecord.rbs.

gems/activerecord/6.0/activerecord.rbs
```diff
@@ -51,6 +51,7 @@ module ActiveRecord
     def self.create!: (**untyped) -> instance
     def self.validate: (*untyped, ?if: condition[instance], ?unless: condition[instance], **untyped) -> void
     def self.validates: (*untyped, ?if: condition[instance], ?unless: condition[instance], **untyped) -> void
+    def self.create_with: (nil | Hash[untyped, untyped]) -> Relation
```
tk0miya, line 54: Could you move this into `ActiveRecord::Base::ClassMethods`, please?
And please copy it onto `_ActiveRecord_Relation_ClassMethods` too.

The rbs_rails gem uses these components and effects the subclasses of ActiveRecord::Base.

dak2: I missed both of them. I did it.
Thanks for your review!

tk0miya (commented):
APPROVE

Thank you for your work!

PR #766, opened 2025-01-10
activemodel: Add ActiveModel::Errors#add to 7.1

tk0miya:
In https://github.com/ruby/gem_rbs_collection/pull/717, the signature of `ActiveModel::Errors#add` was moved to activemodel-x.y.rbs from activemodel-generated.rbs (shared).

Unfortunately, that change forgot to copy the signature to activemodel/7.1.  Therefore, the method is not found if users installs activemodel-7.1 or later.

This copies the signature of `Errors#add` from 7.0.


ksss (commented):
APPROVE

PR #778, opened 2025-01-23
activerecord: Add types for previously_persisted?

dak2:
This method has been added since 7.0

- https://github.com/rails/rails/pull/42389
- https://github.com/rails/rails/blob/7-0-stable/activerecord/CHANGELOG.md


tk0miya (commented):
APPROVE

PR #779, opened 2025-01-25
add: ConnectionHandling#lease_connection

rhiroe:
https://github.com/rails/rails/blob/v7.2.0/activerecord/lib/active_record/connection_handling.rb#L255


ksss (commented):
APPROVE

PR #780, opened 2025-01-25
corrected the type of ConnectionHandling#connects_to

rhiroe:
https://github.com/rails/rails/blob/v8.0.1/activerecord/lib/active_record/connection_handling.rb#L81


ksss (commented):
APPROVE

PR #781, opened 2025-01-29
Add types for active_type gem

kyanagi:
Types for [active_type](https://github.com/makandra/active_type) gem.


kyanagi:
PR updated: The second argument of `ActiveType.cast` has been restricted to subclasses of `ActiveRecord::Base`.

PR #787, opened 2025-02-11
add: sidekiq-pro

rhiroe:
The source code of sidekiq-pro is not publicly available, but I have permission to publish the RBS as long as we explicitly state at the beginning of the file that it is unofficial.

It is generated using `rbs prototype rb`.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -4,6 +4,10 @@ require "pathname"
 require "optparse"
 require 'yaml'
 
+excluded_gems = %w[
+  sidekiq-pro
+]
+
 @all = false
 @verbose = false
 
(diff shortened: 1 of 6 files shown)


rhiroe:
@pocke Since sidekiq-pro cannot be fetched from https://rubygems.org/, the CI is failing. For now, I have excluded it from the tests, but I would appreciate your thoughts on this approach.

bin/test
```diff
@@ -4,6 +4,10 @@ require "pathname"
 require "optparse"
 require 'yaml'
 
+excluded_gems = %w[
+  sidekiq-pro
+]
+
```
tk0miya, line 10: `bin/test` automatically skips running tests if `_test/` directory is not found. So you don't need to manage this list. How about removing the directory?
https://github.com/ruby/gem_rbs_collection/blob/569dd4879c76d8ed869d799a3d2ec916727ffbb2/bin/test#L181-L186

tk0miya:
IMO, this is worthy even if without CI. +1 for merging.

PR #790, opened 2025-02-12
activerecord: Make block optional in Base.find_by_sql

nickjer:
According to [the documention](https://api.rubyonrails.org/classes/ActiveRecord/Querying.html#method-i-find_by_sql) for `ActiveRecord::Querying.find_by_sql` you do not need to define a block when calling the method. But Steep raises the following error...

```
activerecord-generated.rb:43:5: [error] The method cannot be called without a block
│
└ User.find_by_sql("SELECT * FROM users")
       ~~~~~~~~~~~
```

This change should make the block optional.


gems/activerecord/6.0/activerecord-generated.rbs
```diff
@@ -16651,7 +16651,7 @@ module ActiveRecord
     #
     #   Post.find_by_sql ["SELECT title FROM posts WHERE author = ? AND created > ?", author_id, start_date]
     #   Post.find_by_sql ["SELECT body FROM comments WHERE author = :user_id OR approved_by = :user_id", { :user_id => user_id }]
-    def find_by_sql: (untyped sql, ?untyped binds, ?preparable: untyped? preparable) { () -> untyped } -> untyped
+    def find_by_sql: (untyped sql, ?untyped binds, ?preparable: untyped? preparable) ?{ () -> untyped } -> untyped
```
tk0miya, line 16654: Looks good. Could you move this definition to activerecord.rbs, please?
We use that file to manage manually modified definitions.

nickjer: My bad, forgot about the naming convention between the files :sweat_smile:

nickjer: Let me know if I should also remove the large code comment above the method signature. I was unsure about that.

tk0miya (commented):
APPROVE

gems/activerecord/6.0/activerecord-generated.rbs
tk0miya: Good change! It would be better to move both signature and comment :-)

nickjer:
Well, looks like my bad habit of pulling in the latest `master` branch has led to this getting out of sync with the approval. I wonder if I should remove that merge and force push.

tk0miya (commented):
APPROVE!

PR #791, opened 2025-02-12
activesupport: Add types to AS::RangeWithFormat#to_s and AS::NumericWithFormat#to_s

tk0miya:
`#to_s` methods should return String object.

Additionally, I moved the definitions to activesupport-6.0.rbs, not activesupport.rbs because `#to_s` overrides ware marked as deprecated since v7.0, and removed since v7.1.

refs:

* https://github.com/rails/rails/pull/43772
* https://github.com/rails/rails/commit/e420c3380eb2b698a4fe84ed196f914d18f7844a
* https://github.com/soutaro/steep/issues/1478

### DIFF
--- a/gems/activesupport/6.0/activesupport-6.0.rbs
+++ b/gems/activesupport/6.0/activesupport-6.0.rbs
@@ -24,6 +24,131 @@ module ActiveSupport
   end
 end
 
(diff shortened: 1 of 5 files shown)


ksss (commented):
APPROVE

PR #797, opened 2025-02-15
activerecord: Add block support to sum method

euglena1215:
In addition to passing column names, the sum method can also pass blocks. https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/relation/calculations.rb#L84-L94

```rb
users = User.all
users.sum(:age) # Already defined in type definition
users.sum(&:age) # Not defined in type definition
```

This change adds support for the sum method to accept a block.


ksss (commented):
APPROVE

PR #807, opened 2025-02-26
Exclude private API

ksss:
* Remove instance/class_instance/class variable
* Remove private method

As stated in the guidelines, important APIs should be documented, but private APIs should not be included.
https://github.com/ruby/gem_rbs_collection/blob/main/docs/CONTRIBUTING.md#focus-on-the-most-important-part-of-the-api

The Sidekiq Pro documentation does not mention the use of instance variables or private methods.
https://github.com/sidekiq/sidekiq/wiki#sidekiq-pro

It passed validation in my application using sidekiq-pro.


ksss:
@rhiroe Please review this changes.

rhiroe (commented):
APPROVE

PR #814, opened 2025-03-07
fix activerecord sum methods

rhiroe:
When a block is given, it is delegated to Enumerable#sum. See `Enumerable#sum` for reference.

Since it is more type-friendly for the types of `initial_value`, the block's return value, and the method's return value to match, I will accept some imprecision and unify all types as `T`.

When no block is given, the calculate method processes the result, so the return value is not necessarily an Integer.

https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/relation/calculations.rb#L127

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -431,8 +431,8 @@ module ActiveRecord
                    | (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) -> Enumerator[Model, nil]
       def find_in_batches: (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) { (Array[Model]) -> void } -> nil
       def in_batches: (?of: Integer, ?start: Integer, ?finish: Integer, ?load: bool, ?error_on_ignore: bool, ?order: untyped) { (self) -> void } -> nil
-      def sum: (?untyped? column_name) -> Integer
-             | () { (Model) -> Integer } -> Integer
+      def sum: (?untyped? column_name) -> untyped
+             | [T] (?T initial_value) { (Model) -> T } -> T
       def destroy_all: () -> untyped
       def delete_all: () -> untyped
       def update_all: (*untyped) -> untyped
@@ -528,8 +528,8 @@ module ActiveRecord
                    | (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) -> Enumerator[Model, nil]
       def find_in_batches: (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) { (Array[Model]) -> void } -> nil
       def in_batches: (?of: Integer, ?start: Integer, ?finish: Integer, ?load: bool, ?error_on_ignore: bool, ?order: untyped) { (Relation) -> void } -> nil
-      def sum: (?untyped? column_name) -> Integer
-             | () { (Model) -> Integer } -> Integer
+      def sum: (?untyped? column_name) -> untyped
+             | [T] (?T initial_value) { (Model) -> T } -> T
       def destroy_all: () -> untyped
       def delete_all: () -> untyped
       def update_all: (*untyped) -> untyped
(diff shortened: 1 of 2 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -431,8 +431,8 @@ module ActiveRecord
                    | (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) -> Enumerator[Model, nil]
       def find_in_batches: (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) { (Array[Model]) -> void } -> nil
       def in_batches: (?of: Integer, ?start: Integer, ?finish: Integer, ?load: bool, ?error_on_ignore: bool, ?order: untyped) { (self) -> void } -> nil
-      def sum: (?untyped? column_name) -> Integer
-             | () { (Model) -> Integer } -> Integer
+      def sum: (?untyped? column_name) -> untyped
+             | [T] () { (Model) -> T } -> (Integer | T)
```
ksss, line 435: The intent of my comment was to use overloading.
I don't know if it will work.

```suggestion
      def sum: (?untyped? column_name) -> untyped
             | ...
```

ksss (commented):
I welcome the accuracy of type information, but not the **inconvenience**.

It is likely that adjusting the results based on the column type for each model would be the optimal approach, but in most cases, the type would be either `Integer` or `Float`.  

What specific problem does this change solve for you?

rhiroe:
> It is likely that adjusting the results based on the column type for each model would be the optimal approach, but in most cases, the type would be either Integer or Float.

I agree.

> What specific problem does this change solve for you?

The issue I'm facing is that summing Float values results in an Integer.

gems/activerecord/6.0/activerecord.rbs
rhiroe: I didn’t realize that overloading could be used here, but that makes perfect sense. I'll check it out.

rhiroe: It seems to result in an `InvalidOverloadMethodError`. 😢

gems/activerecord/6.0/activerecord.rbs
```diff
@@ -680,8 +682,9 @@ interface _ActiveRecord_Relation[Model, PrimaryKey] # rubocop:disable RBS/Lint/T
                | (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) -> Enumerator[Model, nil]
   def find_in_batches: (?batch_size: Integer, ?start: Integer, ?finish: Integer, ?error_on_ignore: bool) { (Array[Model]) -> void } -> nil
   def in_batches: (?of: Integer, ?start: Integer, ?finish: Integer, ?load: bool, ?error_on_ignore: bool, ?order: untyped) { (self) -> void } -> nil
-  def sum: (?untyped? column_name) -> Integer
-         | () { (Model) -> Integer } -> Integer
+  def sum: (?untyped? column_name) -> untyped
+         | [T] () { (Model) -> T } -> (Integer | T)
+         | [U] (U initial_value) { (Model) -> untyped } -> U
```
ksss, line 687: The return value does not necessarily match `initial_value`.

```rb
User.sum(0) { |user| user.decimal_column } #=> BigDecimal
```

Also, if there are zero records, the type of initial_value seems to be returned.

```rb
User.sum(0.0) { |user| user.decimal_column } #=> 0.0
```

If you value practicality,

```suggestion
         | [T] (untyped initial_value) { (Model) -> T } -> T
```

If you value accuracy,

```suggestion
         | [T, U] (T initial_value) { (Model) -> U } -> (T | U)
```

ksss (commented):
I generally agree with cases that involve a block.  

For cases without a block, let me think about it a bit more.

ksss:
@rhiroe 

Here’s my thought on cases without a block:  

In conclusion, I believe `untyped` is the best choice.  

To express it completely accurately, the `sum` method would need to be overridden for each column.

```rbs
class User < ActiveRecord::Base
  def self.sum: (:id) -> Integer | ...
  def self.sum: (:hight) -> Float | ...
end
```

A simpler approach would be to use `-> (String | Integer | Float | BigDecimal)?`, which would be sufficient.

However, neither seems particularly convenient to me.
Ultimately, annotating it directly in the code seems like the most reasonable approach.  

```rb
sum_of_hight = User.sum(:hight) #: Float
```

Therefore, I think `untyped` is the best option.  

---

For cases with a block and an `initial_value`, I believe one of the methods I proposed would be appropriate. What do you think?

rhiroe:
Thank you for your suggestion. +1.

I believe the type for cases with `initial_value` should be `[T] (untyped initial_value) { (Model) -> T } -> T`. While this may lack precision, it is more type-friendly if the types of `initial_value`, the block's return value, and the method's return value match, and I believe this is how it should be implemented. Therefore, I think it makes sense to prioritize usability over strict accuracy.

Therefore, I’m considering defining the type as follows:

```rbs
def sum: (?untyped? column_name) -> untyped
       | [T] (?T initial_value) { (Model) -> T } -> T
```
At this point, I don’t see the need to differentiate the type based on the presence of `initial_value`. If a concrete issue arises from this approach, we can reconsider it at that time.

This is my current thought on the matter. What do you think?

ksss:
`[T] (?T initial_value) { (Model) -> T } -> T` looks great to me!

ksss (commented):
APPROVE

PR #815, opened 2025-03-08
activesupport: Update return types of `#present?` and `#blank?`

tk0miya:
This updates the return types of `#present?` and `#blank?` to narrow the type of receiver possibly.

Since 1.10, Steep can narrow the type of receiver via the return type of the method when the type of receiver is union.  For example, a receiver typed as will be narrowed to `String` on the `if present?` block, and be narrowed to `String | nil` on the `else` block because `String#present?` returns bool and `NilClass#present?` returns false.

```ruby
var = ... # String | nil
if var.present?
  # var is narrowed to String
else
  # var is narrowed to String | nil
end
```

refs:

* https://github.com/soutaro/steep/wiki/Release-Note-1.10 (unreleased yet)
* https://github.com/soutaro/steep/pull/1497

### DIFF
--- a/gems/activesupport/6.0/activesupport.rbs
+++ b/gems/activesupport/6.0/activesupport.rbs
@@ -462,8 +462,8 @@ end
 
 # active_support/core_ext/object/blank.rb
 class Object
-  def blank?: () -> bool
-  def present?: () -> bool
+  def blank?: () -> false
+  def present?: () -> true
   def presence: () -> self?
 end
 
@@ -475,22 +475,27 @@ end
 
 class NilClass
   def blank?: () -> true
+  def present?: () -> false
 end
 
 class FalseClass
   def blank?: () -> true
+  def present?: () -> false
 end
 
 class TrueClass
   def blank?: () -> false
+  def present?: () -> true
 end
 
 class Array[unchecked out Elem]
   alias blank? empty?
+  def present?: () -> bool
 end
 
 class Hash[unchecked out K, unchecked out V]
   alias blank? empty?
+  def present?: () -> bool
 
   # Returns a string containing an XML representation of its receiver:
   #
@@ -561,6 +566,7 @@ class String
   BLANK_RE: Regexp
   ENCODED_BLANKS: Concurrent::Map[Encoding, Regexp]
   def blank?: () -> bool
+  def present?: () -> bool
 end
 
 class Numeric


gems/activesupport/6.0/activesupport.rbs
```diff
@@ -462,8 +462,8 @@ end
 
 # active_support/core_ext/object/blank.rb
 class Object
-  def blank?: () -> bool
-  def present?: () -> bool
+  def blank?: () -> false
+  def present?: () -> true
```
ksss, line 466: Is my understanding correct that subclasses now **must** override it as needed?

soutaro: Hmm. Having bool return type here would make more sense technically, while I would recommend adding type decls if a subclass has custom implementation.

tk0miya: I think using false here is incorrect, but useful for almost all cases.

I agree that using bool here is more correct and precise. But it means users should define type decls manually for almost of subclasses even if they do not have custom implementation because `Object#present?` always returns false if the class does not have `#empty?`.

For example, an empty class having no own methods returns false for the `#blank?` call.

```
root@288c83a8b21f> bundle exec irb
irb(main):001> require 'active_support/all'
=> true
irb(main):002> class MyClass; end
=> nil
irb(main):003> MyClass.new.blank?
=> false
```

On the other hand, as @ksss said, the subclasses that override `#blank?` or implement `#empty?` must override the return type of `#blank?` and `#present?`.  This must be a trap.


Refs:

* https://api.rubyonrails.org/classes/Object.html#method-i-blank-3F
* https://github.com/rails/rails/blob/741e09930d86d81d76931ca62e8a913e4f8ee417/activesupport/lib/active_support/core_ext/object/blank.rb#L18-L20

tk0miya: It's enough worthy even if we choose bool.

```
optional_val = ...   # SomeObject | nil
if optional_val.present?
  # narrowed to SomeObject
else
  # type narrowing does not work here because SomeObject#present? returns a bool.
end

if optional_val.blank?
  # type narrowing does not work here
else
  # narrowed to SomeObject
end
```

I feel this is not perfect, but acceptable.

ksss: I was concerned that knowledge of the importance of the `#empty` implementation is required and that there is a possibility of type misdetection.
However, since we are prioritizing practicality, Let's give it a try.

ksss (commented):
APPROVE

gems/activesupport/6.0/activesupport.rbs
tk0miya: Thank you for your comment. Okay, let's go with this.

PR #829, opened 2025-03-18
Fix `RBS::NoSuperclassFoundError`

ksss:
Our application uses **activeadmin** but not **devise**.
When we run `rbs validate` on this application, it fails with `RBS::NoSuperclassFoundError`.


ksss:
@euglena1215 FYI

euglena1215:
Ah, sorry. Thanks for the fix!

PR #840, opened 2025-03-27
csv: Add signature for CSV.open

KitaitiMakoto:
Hello,

I added signature for `CSV.open`.
Can you review it?

Thank you.

### DIFF
--- a/gems/csv/3.3/_test/test.rb
+++ b/gems/csv/3.3/_test/test.rb
@@ -2,7 +2,6 @@
 # It is type checked by `steep check` command.
 
 require "csv"
-require "tmpdir"
 
 csv = CSV.new("header1,header2\nrow1_1,row1_2")
 csv.headers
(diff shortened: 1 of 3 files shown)


gems/csv/3.3/manifest.yaml
```diff
@@ -1,3 +1,4 @@
 dependencies:
   - name: forwardable
   - name: stringio
+  - name: tmpdir
```
ksss, line 4: It does not appear to be necessary.

gems/csv/3.3/csv.rbs
```diff
@@ -1700,6 +1700,109 @@ class CSV < Object
...
+                  ?encoding: Encoding | String,
+                  **untyped options) { (instance) -> void } -> instance
+               | (String | _ToPath | IO filename_or_io,
+                  ?String mode,
+                  ?newline: untyped,
+                  ?invalid: nil | :replace,
+                  ?undef: :replace,
+                  ?replace: String,
+                  ?encoding: Encoding | String,
+                  **untyped options) -> instance
```
ksss, line 1804: The block call appears to be just optional.
If so, we can omit the signature as follows.

```suggestion
  def self.open: (String | _ToPath | IO filename_or_io,
                  ?String mode,
                  ?newline: untyped,
                  ?invalid: nil | :replace,
                  ?undef: :replace,
                  ?replace: String,
                  ?encoding: Encoding | String,
                  **untyped options) ?{ (instance) -> void } -> instance
```

gems/csv/3.3/csv.rbs
```diff
@@ -1700,6 +1700,109 @@ class CSV < Object
...
+  #   csv # => #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
+  # Output:
+  #   #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
+  # ---
+  #
+  # Raises an exception if the argument is not a \String object or \IO object:
+  #   # Raises TypeError (no implicit conversion of Symbol into String)
+  #   CSV.open(:foo)
+  #
+  def self.open: (String | _ToPath | IO filename_or_io,
```
ksss, line 1789: Can we support StringIO?

> Using a \StringIO:
>   csv = CSV.open(string_io) {|csv| p csv}
>  csv # => #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
> Output:
>    #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">

gems/csv/3.3/_test/test.rb
```diff
@@ -14,5 +15,68 @@
...
+
+  path6 = File.join(tmpdir, "example6.csv")
+  CSV.open(path6, "w", headers: :first_row)
+  File.open(path6, "w") do |io|
+    CSV.open(io, "w", headers: :first_row) { |csv| csv << ["a", "b", "c"] }
+  end
+  File.open(path6, "w") do |io|
+    CSV.open(io, "w", headers: :first_row)
+  end
+end
```
ksss, line 79: It is not necessary to cover every combination of options.
This is because this code is not executed on the test and we are not testing the behavior.
We only need the patterns shown by the overloads and the union combinations of the main arguments.

KitaitiMakoto: @ksss 
Let me make it clear.

* After https://github.com/ruby/gem_rbs_collection/pull/840/commits/96a05d0e76a42f41e19aed2a024d76ea0d1d386a is committed, there's a single overload,
* there are three main arguments: `filename_or_io`, `options` and a block

If so, we should test 16 patterns, right?: 4 patterns of `filename_or_io` (`String`, `_ToPath`, `IO` and `StringIO`) x 2 patterns of `options` (no options and some options) x 2 patterns of the block (no block and a block) = 16

In addition, I want to increase patterns. I think we can split `options` into three groups:

* `CSV.open`-specific: `:newline`, `:invalid`, `:undef` and `:replace`,
* common options for class methods: `:encoding` (maybe `:external_encoding` and `:internal_encoding` are true but skipping for now because it's not documented), and
* Options for Generating

So, I want count 4 patterns for `options` (three for the above and one for none).

The reason I think three groups is because the last two groups can be defined using RBS's `type` notation so that other methods can use it and those might change and test patterns might became not enough in the future refactoring of RBS, though it might violate the YAGNI principle. How do you think?

ksss: - No behavioral testing is required. A type test is required. We would like about a major sample code.
- These tests are temporary and will all be removed when we eventually rewrite them as unit tests for the gem itself.
- The more tests, the more administrative costs. I would prefer minimal testing.

gems/csv/3.3/manifest.yaml
ksss: A metadata.yaml is required for types used only for testing.

Here is a good examples.
https://github.com/ruby/gem_rbs_collection/blob/main/gems/redis/4.2/_test/metadata.yaml

gems/csv/3.3/_test/test.rb
KitaitiMakoto: @ksss Thank you for description. I realized and fixed it. Is it what you mean?

gems/csv/3.3/manifest.yaml
KitaitiMakoto: Yeah, I misunderstood that `bin/test` tests Ruby code behaviors. If so, we have needed `tmpdir` to generate files for testing. But now I understood it's not required.

ksss:
APPROVE

ksss (commented):
APPROVE

KitaitiMakoto:
Thank you for leading me to proper patch.

PR #846, opened 2025-04-12
refactor: replace `Object#name` with `(String | Symbol)` in ActiveSupport

rhiroe:
Because `Object::name` type is deprecated.

### DIFF
--- a/gems/activesupport/6.0/activesupport.rbs
+++ b/gems/activesupport/6.0/activesupport.rbs
@@ -360,22 +360,22 @@ module ActiveSupport
   module Tryable : BasicObject
     def try: [T] () { (self) -> T } -> T # When yield self
            | [T] () { () -> T } -> T # When instance_eval
-           | (Object::name method_name, *untyped args) ?{ (*untyped) -> untyped } -> untyped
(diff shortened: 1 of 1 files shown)


ksss (commented):
I agree with removing `name`, but my opinion on using `interned` is the same as [here](https://github.com/ruby/gem_rbs_collection/pull/510#issuecomment-1915923966).

rhiroe:
I’m not particularly attached to using interned. I'm planning to replace all of them with `(String | Symbol)`. I added the TODO comment with good intentions, but it seems it wasn’t necessary after all.

On the other hand, things that would cause problems if used elsewhere shouldn't be defined at the top level. It's hard to agree with the idea of defining something at the top level while saying "I doubt anyone would actually used it."

ksss (commented):
I found the owner's opinion on `interned`.
https://github.com/ruby/rbs/pull/1499#issuecomment-1707865524

I believe there are areas where use is appropriate.
But, keeping the scope of the PR focused would help avoid the risk of the discussion becoming too broad.

ksss (commented):
APPROVE

PR #850, opened 2025-05-08
activerecord: Update callback types to accept CallbackObject

Okunichiyou:
### Background
In the current type definitions for ActiveRecord, callback objects could not be properly accepted as arguments for lifecycle callbacks. This limitation made it difficult to use callback objects in a type-safe manner, even though they are supported by Rails.

### Changes
This PR updates the type definitions to allow callback objects to be passed to the lifecycle callbacks listed in the Rails documentation:  
[Active Record Callbacks - Available Callbacks](https://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks).  
Only the callbacks explicitly listed in the documentation have been addressed in this update.

### How to Verify
To confirm the current issue and verify the changes, please refer to the following repository:  
[https://github.com/Okunichiyou/rbs-pra/tree/test1](https://github.com/Okunichiyou/rbs-pra/tree/test1)

This repository demonstrates the type error that occurs when using callback objects with the current type definitions.

1. Install dependencies:
   ```bash
   bundle install
   ```

2. Run RSpec tests:
   ```bash
   bundle exec rspec spec/models/comment_spec.rb
   ```
   This confirms that the callback object is correctly accepted in Rails.

3. Perform type checking:
   ```bash
   bundle exec steep check
   ```
   This reproduces the type error caused by the current type definitions.

### DIFF
--- a/gems/activerecord/6.0/_test/activerecord.rb
+++ b/gems/activerecord/6.0/_test/activerecord.rb
@@ -1,0 +1,49 @@
+class TestCallbackObject < ActiveRecord::Base
+  before_validation ::Callbacks::ClassCallback
+  after_validation ::Callbacks::ClassCallback
+  before_save ::Callbacks::ClassCallback
+  before_create ::Callbacks::ClassCallback
+  after_create ::Callbacks::ClassCallback
+  after_save ::Callbacks::ClassCallback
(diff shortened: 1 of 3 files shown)


ksss (commented):
APPROVE

PR #858, opened 2025-05-17
add: ActiveRecord::Querying#sole

rhiroe:
## Background
Add type definitions for `ActiveRecord::Querying#sole`, introduced in ActiveRecord 7.0.

## Changes
Move the following interfaces from `gems/activerecord/6.0/activerecord.rbs` to version-specific files so that changes apply only for ActiveRecord 7.0 and above:
- `_ActiveRecord_Relation`
- `_ActiveRecord_Relation_ClassMethods`

cfa0addb7dcc44b6b03c91f48c800cafd22e5e8f

Add type signatures for the following methods in all files targeting ActiveRecord 7.0 and later:
- `ActiveRecord::Base::ClassMethods#sole`
- `ActiveRecord::Relation::Methods#sole`

And add method definitions for sole in the moved interfaces only for versions 7.0 and above:
- `_ActiveRecord_Relation#sole`
- `_ActiveRecord_Relation_ClassMethods#sole`

3de1dfb23bb60ec8797f9edf14a3ddf85746fe97

Add corresponding tests.
Since there was no test file for 8.0, I copied the test from 7.2.


ksss (commented):
APPROVE

PR #859, opened 2025-05-18
Fix type signature for `ActiveRecord::Base.validate` (to match official Ruby documentation)

yuzisee:
According to https://guides.rubyonrails.org/v6.0/active_record_validations.html#combining-validation-conditions

the official documentation says you can have
`if: [Proc.new { |c| c.market.retail? }, :desktop?],`
so, RBS signatures should reflect that.

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -49,7 +49,7 @@ module ActiveRecord
     def self.transaction: [T] (?requires_new: boolish, ?isolation: (:read_uncommitted | :read_committed | :repeatable_read | :serializable)?, ?joinable: boolish) { () -> T } -> T
     def self.create: (**untyped) -> instance
     def self.create!: (**untyped) -> instance
-    def self.validate: (*untyped, ?if: condition[instance], ?unless: condition[instance], **untyped) -> void
+    def self.validate: (*untyped, ?if: condition[instance] | Array[condition[instance]], ?unless: condition[instance] | Array[condition[instance]], **untyped) -> void
     def self.validates: (*untyped, ?if: condition[instance], ?unless: condition[instance], **untyped) -> void
 
     # callbacks


tk0miya (commented):
LGTM. Could you check the errors in CI? I'll merge this after CI passed.

gems/activerecord/6.0/_test/activerecord.rb
```diff
@@ -7,6 +7,26 @@ class TestCallbackObject < ActiveRecord::Base
...
+
+  def local_condition2
+    [true, false].sample
+  end
+
+  def local_condition3
+    [true, false].sample
+  end
+
+  validate :custom_validation, on: :update, unless: [:local_condition1, proc { |this_rec| this_rec.local_condition2 }, ->(my_rec) { my_rec.local_condition3 }]
```
yuzisee, line 23: Re: https://github.com/ruby/gem_rbs_collection/pull/859#pullrequestreview-2848906133

> LGTM. Could you check the errors in CI? I'll merge this after CI passed.

Thanks! The `steep check` is succeeding, but `steep stats` is complaining:
```
⋮
  No type error detected. 🫖
  2 method calls with untyped receiver detected from activerecord.rb. Please assign the expected type to the receivers
⋮
```

But if `->(my_rec) { my_rec.local_condition3 }` is matching `^(T) [self: T] -> boolish` then **my_rec** shouldn't be considered _TypeInference::MethodCall::Untyped_ right? It should be that `[self: T]` is a https://github.com/ruby/rbs/blob/master/docs/syntax.md#self-type-binding

Meanwhile, if `proc { |this_rec| this_rec.local_condition2 }` matches `Proc` then there's no way to set a type for **this_rec** right?

yuzisee: @tk0miya do you think I should try to use https://github.com/soutaro/steep/blob/master/manual/annotations.md here? I'm not sure how else to set the type of the receiver inside the `unless:` block while still keeping the syntax from the original Rails documentation.

tk0miya:
Thank you for your mention. I sincerely apologize. It completely slipped my mind.

I'm not sure why you added test cases to ActiveRecord in spite of modifying ActiveModel.
Is it better to move them to ActiveModel?

tk0miya:
LGTM for this, but I could not approve this. On my local, reviewing UI is not displayed to me.

@ksss Can you approve this?

ksss (commented):
APPROVE

ksss:
@yuzisee @tk0miya Thanks!

tk0miya:
@ksss thanks!

PR #861, opened 2025-05-19
feat(paranoia): add replacements for untyped

casaper:
- type for optional options of `ActiveRecord::Base.acts_as_paranoid`
- fixed wrong or missing response types added to `ActiveRecord::Base` by paranoia gem
- Typing of more methods in paranoia gem models
- Included the class methods into the module Paranoia::Query with `ClassMethods[untyped, untyped]` to prevent my changes braking existing usages of this sig
- included instance methods into Paranoia with `InstanceMethods[untyped]` to prevent braking current sig usages
- Added some more tests


casaper:
/merge 

Can I? Seriously?

Issue #788, opened 2025-02-11
Proposal for Managing Rails Repositories


ksss:
Dear Maintainers of Rails Repositories(@tk0miya, @ydah, @hibariya, @Little-Rubyist).

Currently, Rails repositories follow a method where v6 signatures are inherited, and differences are added incrementally. While this approach keeps management costs low in the short term, I believe it has several issues:

- Prototype-based signatures are often excessive, leading to a decrease in type-checking performance.
- As new versions are introduced, the differences from the base version become larger, reducing sustainability.
- It becomes impossible to stop maintaining v6.

## Proposal

I propose discontinuing the inheritance of symbolic links from a specific version and managing only the necessary definitions manually.

Specifically, I suggest keeping the differential definitions while not inheriting *-generated.rbs, thereby minimizing the definitions. This approach aims to maintain sustainability without compromising performance.

For minor version differences, symbolic links may still be acceptable. However, for major version differences, I believe symbolic links should not be inherited.

What do you think?
I would appreciate your feedback.

tk0miya:
We have six versions for activerecord gem (6.0, 6.1, 7.0, 7.1, 7.2, and 8.0). It's too much and hard to maintain. So, +1 for changing maintenance policy.

According to the maintenance policy of Rails, 6.x becomes EOL now. But 7.0+ is still supported. How about dropping symlinks to 6.x versions?
https://rubyonrails.org/maintenance

In addition to this, how about merging minor versions into one version? IMO, 7.0, 7.1, and 7.2 can be merged. Then there are only 2 or 3 versions we have to maintain at the same time.

Little-Rubyist:
Thank you for your suggestion.
I’m +1 on removing the symbolic link for the major version.
As for the integration of minor versions, I’m concerned about potential exceptions. I haven’t checked if this has happened in the past, but I feel like there could be cases where unintended nullable get fixed.

tk0miya:
+1. Let's follow the upstream's policy.

armstnp:
+1 - started running into the issues caused by this w/ ActiveStorage's specs.

PR #863, opened 2025-05-21
Narrow down supported versions

ksss:
Dear Maintainers of Rails Repositories(@tk0miya, @ydah, @hibariya, @Little-Rubyist).
Please review this PR.

I have responded about activerecord to implement https://github.com/ruby/gem_rbs_collection/issues/788.

# Point of review

Please see `gems/activerecord/README.md`.
If this text allows, it will be expanded to other rails gems.
Other gems do not require review.

## Supported Versions

* 8.0 and 7.2 are the targets.
* Simply copied the original file instead of symlink.
* Some of the test files were copied from 6.0.

## Unsupported Versions

* 7.1, 7.0, 6.1 and 6.0 are the targets.
* Deleted test.
* Excluded from rubocop.
* Symlink was left in place.


tk0miya:
Ah, sorry for my confusion. Please forget my last comment. I've never had permission to approve in GitHub.
The only thing I can do is approve via "APPROVE" comment, not via the review interface.

PR #864, opened 2025-05-21
Allow star(`*`) in scalar value of rubocop.yml

ksss:
# Problem

I would like to exclude files that are no longer supported by RuboCop, but star (`*`) cannot be used in path specifications. (e.g. `"gems/activerecord/7.2/**/*.rbs"`)
Since the target includes many old versions of various Rails-related gems, listing all files individually increases maintenance costs.

# Proposal

I propose allowing the use of `*` in path specifications.
The `*` character is currently disallowed due to security concerns, as it can be used for YAML aliasing. However, since `YAML.safe_load` is already in use and does not permit aliases, this particular concern seems to be already addressed.

To further safeguard against unintentionally enabling alias functionality after allowing `*`, we have also added a guard clause.

# Ref

https://github.com/ruby/gem_rbs_collection/actions/runs/15153815288/job/42604695763?pr=863
```
Run ruby .github/workflows/pr_bot/strict_rubocop_config.rb
.github/workflows/pr_bot/strict_rubocop_config.rb:66:in `validate_rubocop_yml!': Invalid characters in gems/activerecord/.rubocop.yml:7:30. (InvalidCharacters)
    - "gems/activerecord/7.2/**/*.rbs"
                             ^

	from .github/workflows/pr_bot/strict_rubocop_config.rb:9[5](https://github.com/ruby/gem_rbs_collection/actions/runs/15153815288/job/42604695763?pr=863#step:4:6):in `block in <main>'
	from .github/workflows/pr_bot/strict_rubocop_config.rb:[8](https://github.com/ruby/gem_rbs_collection/actions/runs/15153815288/job/42604695763?pr=863#step:4:9)9:in `each'
	from .github/workflows/pr_bot/strict_rubocop_config.rb:89:in `<main>'
```

### DIFF
--- a/.github/workflows/pr_bot/strict_rubocop_config.rb
+++ b/.github/workflows/pr_bot/strict_rubocop_config.rb
@@ -17,7 +17,7 @@
 require 'open3'
 require 'yaml'
 
-DISALLOWED_CHARACTERS_RE = /[^a-zA-Z0-9\s_\/#:'".,\-]/
+DISALLOWED_CHARACTERS_RE = /[^a-zA-Z0-9\s_\/#:'".,\-*]/
 
 class InvalidCharacters < StandardError
   def initialize(path:, match:)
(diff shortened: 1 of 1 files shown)


pocke (approved):
Looks good, thanks!

PR #876, opened 2025-06-09
parser: Add types for Parser::Source

kozy4324:
- Add `#text` to `Parser::Source::Comment` class
- Add `#node` and `#expression` to `Parser::Source::Map` class
- Add some methods to `Parser::Source::Range`
- Write tests for the added methods
    - Add `#parse_with_comments` to `Parser::Base` class


pocke (approved):
Thanks!

PR #882, opened 2025-06-14
actioncontroller: fix render_to_string method signature

rhiroe:
Arguments can be passed, and the block is optional.

### DIFF
--- a/gems/actionpack/6.0/actionpack-generated.rbs
+++ b/gems/actionpack/6.0/actionpack-generated.rbs
@@ -511,18 +511,6 @@ module AbstractController
     # sticks the result in <tt>self.response_body</tt>.
     def render: (*untyped args) { () -> untyped } -> untyped
 
-    # Raw rendering of a template to a string.
-    #
-    # It is similar to render, except that it does not
-    # set the +response_body+ and it should be guaranteed
-    # to always return a string.
-    #
-    # If a component extends the semantics of +response_body+
-    # (as ActionController extends it to be anything that
-    # responds to the method each), this method needs to be
-    # overridden in order to still return a string.
-    def render_to_string: (*untyped args) { () -> untyped } -> untyped
-
     # Performs the actual template rendering.
     def render_to_body: (?::Hash[untyped, untyped] options) -> nil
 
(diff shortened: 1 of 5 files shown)


gems/actionpack/6.0/actionpack-generated.rbs
```diff
@@ -2748,9 +2736,6 @@ module ActionController
 
     def render: (*untyped args) -> untyped
 
-    # Overwrite render_to_string because body can now be set to a Rack body.
-    def render_to_string: () -> untyped
```
tk0miya, line 2752: Could you move this entry instead of removal?

ref: https://github.com/rails/rails/blob/v8.0.2/actionpack/lib/action_controller/metal/rendering.rb#L174

gems/actionpack/6.0/actionpack-generated.rbs
```diff
@@ -511,18 +511,6 @@ module AbstractController
...
-    #
-    # It is similar to render, except that it does not
-    # set the +response_body+ and it should be guaranteed
-    # to always return a string.
-    #
-    # If a component extends the semantics of +response_body+
-    # (as ActionController extends it to be anything that
-    # responds to the method each), this method needs to be
-    # overridden in order to still return a string.
-    def render_to_string: (*untyped args) { () -> untyped } -> untyped
```
tk0miya, line 524: It seems this method takes a block, but it never uses it. So it's okay to make it optional.
https://github.com/rails/rails/blob/v8.0.2/actionpack/lib/abstract_controller/rendering.rb#L44

gems/actionpack/6.0/actionpack-generated.rbs
```diff
@@ -511,18 +511,6 @@ module AbstractController
     # sticks the result in <tt>self.response_body</tt>.
     def render: (*untyped args) { () -> untyped } -> untyped
```
tk0miya, line 512: How about making the block argument optional, too?

tk0miya (commented):
LGTM with nits. Could you change the type of `#render`?

gems/actionpack/6.0/actionpack-generated.rbs
rhiroe: As for `render`, in my use case I haven't encountered any issues with the current signature. I also haven’t investigated whether the block for `render` is optional. I don't think I'm the right person to make changes here.

gems/actionpack/6.0/actionpack-generated.rbs
rhiroe: Addressed in this commit 3f1ba37.

ksss (commented):
APPROVE

But, We will not review or test v6.0 in accordance with our maintenance policy.

PR #885, opened 2025-06-18
Fix ActiveRecord read_attribute type

babashunya:
## Overview
When calling the `read_attribute` in ActiveRecord, `The method cannot be called without a block` error occurs.

```
activerecord-generated.rb:37:5: [error] The method cannot be called without a block
│
└ user.read_attribute(:name)
```

The block argument was required by the current type definition, so I made it optional.

## References
- read_attribute
  - https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Read.html


ksss (commented):
@babashunya Thank you for contributing.

Please move signature to activerecord.rbs if you change signature on acitverecord-generated.rbs.
We should document this implicit rule....

babashunya:
@ksss 
Thank you. I moved signature to activerecord.rbs.

PR #898, opened 2025-06-30
activerecord: Add types for `from` to ActiveRecord ClassMethods

dak2:
Currently, the `from` method is defined for the `ActiveRecord::QueryMethods` class, but not for the class method of the ActiveRecord type.

Therefore, `User.from(something)` is a type error.

However, `from` is actually used to chain model classes to retrieve records.

The related file in `rails/rails` has a comment about this. See the link below.

ref. https://github.com/rails/rails/blob/a11f0a63673d274c59c69c2688c63ba303b86193/activerecord/lib/active_record/relation/query_methods.rb#L1359-L1368


ksss (commented):
APPROVE

PR #914, opened 2025-07-15
`I18n.with_locale` should return block result

ksss:
Our application has the following implementation, which has lost type information.

```rb
def foo
  I18n.with_locale(:ja) do
    { bar: }
  end
end
```


ksss:
@takahashim How can I help you?
If it is difficult to continue maintenance, I will take over.

ksss:
Since I didn't receive a response for about a month, I added myself to the review and merged it with administrator privileges.

PR #932, opened 2025-08-26
Bugfix and regressiontest for `Engine.routes`

yuzisee:
Add a unit test to protect against future issues such as https://github.com/ruby/gem_rbs_collection/pull/931#discussion_r2298652222

RouteSet `def url_helpers` modelled after:
```
>>> Rails.application.routes.method(:url_helpers).source_location
["/usr/local/bundle/gems/actionpack-6.0.6.1/lib/action_dispatch/routing/route_set.rb", 478]
```

### DIFF
--- a/gems/railties/6.0/_test/test.rb
+++ b/gems/railties/6.0/_test/test.rb
@@ -2,7 +2,7 @@ Rails.application.routes.draw do
   resources :test1_models
 end
 
-raise <<~HEREDOC unless Rails.application.routes.url_helpers.polymorphic_path(:new_test1_model).instance_of?(String)
+raise <<~HEREDOC unless Rails.application.routes.url_helpers.polymorphic_path(:new_test1_model) == '/test1_models/new'
   But it should work, since:
   >>> Rails.application.routes.method(:url_helpers).source_location
   ["/usr/local/bundle/gems/actionpack-6.0.6.1/lib/action_dispatch/routing/route_set.rb", 478]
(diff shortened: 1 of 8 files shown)


gems/railties/6.0/_test/test.rb
yuzisee, line 1: The test is confirmed to have reproduced the issue
https://github.com/ruby/gem_rbs_collection/pull/932/commits/9cc413910bb275b02f2bd9a52f641fbbf61d4dac

gems/railties/6.0/_test/test.rb
```diff
@@ -0,0 +1,10 @@
+Rails.application.routes.draw do
+  resources :test1_models
+end
+
+raise <<~HEREDOC unless Rails.application.routes.url_helpers.polymorphic_path(:new_test1_model) == '/test1_models/new'
```
yuzisee, line 5: Thanks @ksss

Re: https://github.com/ruby/gem_rbs_collection/pull/932#pullrequestreview-3153546796

> Since `*-generated.rbs` files are intended to remain as generated code, please refrain from adding or modifying type definitions in them—only deletions should be made. For example, if you need to modify `actionpack-generated.rbs`, please move the method to `actionpack.rbs`.

Understood. Done!

> In CI, test code is not executed by the interpreter; only static checking with steep is performed. Will that be sufficient to resolve the issue?

Yes, I only wrote it as `raise … == 'a_string_here'` to make it look like a real file. The comparison operator will ensure that `steep` can only succeed if the type returned by `Rails.application.routes.url_helpers.polymorphic_path(:new_test1_model)` is also a string, so it's all good 🙂

ksss (commented):
APPROVE

PR #935, opened 2025-09-12
Fix type definition for `ActionDispatch::Routing::RouteSet#url_helpers`

nissy325:
Fix the type definition for `ActionDispatch::Routing::RouteSet#url_helpers` to `Module`

references

- https://github.com/rails/rails/blob/9204eb520c2784ca7a1da9a4884aad21c59088fd/actionpack/lib/action_dispatch/routing/route_set.rb#L518-L529
- https://github.com/rails/rails/blob/28bb76d3efc39b2ef663dfe2346f7c2621343cd6/actionpack/lib/action_dispatch/routing/route_set.rb#L478-L481


ksss (commented):
APPROVE

PR #939, opened 2025-10-19
Fix kaminari-activerecord relation type

okuramasafumi:
Previously, we needed to add our own handwritten type to AR model.

```rbs
class Talk < ApplicationRecord
  class ActiveRecord_Relation
    def page: (untyped) -> self
    def per: (Integer) -> self
  end
end
```

This is fixed by this Pull Request.

### DIFF
--- a/gems/kaminari-activerecord/1.2/kaminari-activerecord.rbs
+++ b/gems/kaminari-activerecord/1.2/kaminari-activerecord.rbs
@@ -25,10 +25,9 @@ module ActiveRecord
   end
 
   class Relation
-    module Methods[Model, PrimaryKey]
-      include Kaminari::ActiveRecordRelationMethods
-      include Kaminari::PageScopeMethods
-      def page: (?_ToI num) -> self
-    end
+    include Kaminari::ActiveRecordRelationMethods
+    include Kaminari::PageScopeMethods
+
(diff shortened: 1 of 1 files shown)


Little-Rubyist:
I’m checking this because I think we might also need to update the tests. :loading:

okuramasafumi:
OK, so now I noticed that what is a problem.
The example below works:

```ruby
Talk.page(1).per(10)
```

But this doesn't:

```ruby
Talk.all.page(1).per(10)
```

I believe it's because `all` returns a type which we don't expect. I have not investigated it well yet, though.

ksss:
`page` method of kaminari can be dynamically changed by configuration.

https://github.com/kaminari/kaminari/blob/master/README.md#changing-page_method_name

If we fix the name here, it may be useful for many applications that have not changed their settings, but it may cause confusion for some applications that have changed their settings.
For more correct typing, I recommend writing types that match the configuration on the application side.

okuramasafumi:
> page method of kaminari can be dynamically changed by configuration.

Right, but if it's not common, adding `page` method makes sense. Writing types manually could be optional.

ksss:
In RBS, we cannot redefine the same method name, nor can we invalidate existing definitions.
Optional means would require advanced hacks.

PR #941, opened 2025-10-21
actionview: Fix form tags without blocks and some block types

felixefelip:
Some tags like `label_tag` of form were with required block, but as in the Rails docs it's possible to use it without a block
image

`label_tag` implementation:
image

I also added some block types like `ActionView::Helpers::Tags::CollectionCheckBoxes::CheckBoxBuilder` in `collection_check_boxes`


felixefelip:
> @felixefelip Thank you for requesting. I am positive about accepting this change.
> 
> But, we want to keep gems/actionview/6.0/**actionview-generated.rbs** as an auto-generated type, so if we edit it, we would like to move the type to gems/actionview/6.0/**actionview.rbs**.

@ksss Thank you for the review.

I have moved now the types changed to `gems/actionview/6.0/actionview.rbs`.

ksss (commented):
APPROVE

PR #1020, opened 2026-03-29
rubocop: Add types of rubocop

tk0miya:
* RuboCop::Cop::Base#on_*
* RuboCop::Cop::ConfigurableEnforcedStyle#*_style_detected
* RuboCop::Ext::Comment

refs:

* https://github.com/rubocop/rubocop/blob/v1.57.0/lib/rubocop/cop/base.rb#L137
* https://github.com/rubocop/rubocop/blob/v1.57.0/lib/rubocop/cop/mixin/configurable_enforced_style.rb
* https://github.com/rubocop/rubocop/blob/v1.57.0/lib/rubocop/ext/comment.rb

### DIFF
--- a/gems/rubocop/1.57/rubocop.rbs
+++ b/gems/rubocop/1.57/rubocop.rbs
@@ -35,6 +35,9 @@ module RuboCop
 
       attr_reader config: Config
 
+      def on_new_investigation: () -> void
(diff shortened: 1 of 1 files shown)


kozy4324 (commented):
APPROVE

PR #1035, opened 2026-05-01
rubocop: Fix the patch for Parser::Source::Comment

tk0miya:
RuboCop::Ext::Comment should be included into Parser::Source::Comment, not RuboCop::AST::Comment.  The current patch is incorrect.  This fixes the patch to be applied to the correct class.

ref:

* #1020
* https://github.com/rubocop/rubocop/blob/master/lib/rubocop/ext/comment.rb

### DIFF
--- a/gems/rubocop/1.57/rubocop.rbs
+++ b/gems/rubocop/1.57/rubocop.rbs
@@ -193,16 +193,20 @@ module RuboCop
   end
 end
 
-# Patch to RuboCop::AST::*
+# Patch to RuboCop::AST::* and Parser::Source::Comment
 
 module RuboCop
   module AST
-    class Comment
-      include Ext::Comment
-    end
-
     class ProcessedSource
       include Ext::ProcessedSource
+    end
+  end
+end
+
+module Parser
+  class Source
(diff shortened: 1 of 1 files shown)


gems/rubocop/1.57/rubocop.rbs
```diff
@@ -193,16 +193,20 @@ module RuboCop
...
-    end
-
     class ProcessedSource
       include Ext::ProcessedSource
     end
   end
 end
+
+module Parser
+  class Source
```
kozy4324, line 207:
```suggestion
  module Source
```

The test output shows:

```
Duplicated declaration: ::Parser::Source (RBS::DuplicatedDeclarationError)
```

`::Parser::Source` is already defined as a `module`:

https://github.com/ruby/gem_rbs_collection/blob/3099a933aa5114fb44c04b92c63da49df624123c/gems/parser/3.2/parser.rbs#L53

Changing this should resolve the issue.

kozy4324 (commented):
@tk0miya 
Sorry, I just noticed this. It looks like this may need a fix.

PR #1037, opened 2026-05-18
faraday: Add logging formatter signatures

rossta:
As documented in https://lostisland.github.io/faraday/#/middleware/included/logging?id=customize-the-formatter

cc reviewer @yykamei 

Thank you!

### DIFF
--- a/gems/faraday/2.5/faraday.rbs
+++ b/gems/faraday/2.5/faraday.rbs
@@ -225,6 +225,12 @@ module Faraday
   end
 
   class Response
+    class Logger < Middleware
+      def initialize: (untyped app, ?untyped logger, ?::Hash[Symbol, untyped] options) ?{ (Logging::Formatter formatter) -> void } -> void
+      def call: (Env env) -> untyped
+      def on_complete: (Env env) -> void
+    end
+
     def status: () -> Integer
     def headers: () -> untyped
     def body: () -> String
@@ -262,6 +268,42 @@ module Faraday
     def initialize: (?untyped? app, ?::Hash[untyped, untyped] options) -> void
     def call: (Env env) -> untyped
     def close: () -> void
+  end
+
+  module Logging
+    class Formatter
+      extend Forwardable
+
+      DEFAULT_OPTIONS: ::Hash[Symbol, untyped]
+
+      @logger: untyped
+      @options: ::Hash[Symbol, untyped]
+      @filter: ::Array[[untyped, untyped]]
+
+      def initialize: (logger: untyped, options: ::Hash[Symbol, untyped]) -> void
+
+      def debug: (?untyped progname) ?{ () -> untyped } -> untyped
+      def info: (?untyped progname) ?{ () -> untyped } -> untyped
+      def warn: (?untyped progname) ?{ () -> untyped } -> untyped
+      def error: (?untyped progname) ?{ () -> untyped } -> untyped
+      def fatal: (?untyped progname) ?{ () -> untyped } -> untyped
+
+      def request: (Env env) -> void
+      def response: (Env env) -> void
+      def filter: (untyped filter_word, untyped filter_replacement) -> void
+
+      private
+
+      def dump_headers: (untyped headers) -> String
+      def dump_body: (untyped body) -> String
+      def pretty_inspect: (untyped body) -> String
+      def log_headers?: (String | Symbol type) -> boolish
+      def log_body?: (String | Symbol type) -> boolish
+      def apply_filters: (untyped output) -> String
+      def log_level: () -> Symbol
+      def log_headers: (String type, untyped headers) -> untyped
+      def log_body: (String type, untyped body) -> untyped
+    end
   end
 
   module MiddlewareRegistry
@@ -329,5 +371,8 @@ module Faraday
     attr_accessor response_body: untyped
 
     alias body request_body
+
+    def []: (untyped key) -> untyped
+    def []=: (untyped key, untyped value) -> untyped
   end
 end
--- a/gems/faraday/2.5/_test/test_logging_formatter.rb
+++ b/gems/faraday/2.5/_test/test_logging_formatter.rb
@@ -1,0 +1,22 @@
+class MyFormatter < Faraday::Logging::Formatter
+  def request(env)
+    # Build a custom message using `env`
+    info('Request') { 'Sending Request' }
+  end
+
+  def response(env)
+    # Build a custom message using `env`
+    info('Response') { 'Response Received' }
+  end
+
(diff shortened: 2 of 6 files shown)


gems/faraday/2.5/faraday.rbs
```diff
@@ -264,6 +270,42 @@ module Faraday
     def close: () -> void
   end
 
+  module Logging
+    class Formatter
```
yykamei, line 274: I think these return types should be `void` rather than `untyped`.
In the actual implementation, `request`, `response`, `exception`, and `filter` are side-effect-oriented methods: they write logs or update the formatter's filter list, and their return values are not part of the public contract.

```rbs
def request: (Env env) -> void
def response: (Env env) -> void
def exception: (untyped exc) -> void
def filter: (untyped filter_word, untyped filter_replacement) -> void
```

The test signature in _test/test_logging_formatter.rbs should probably use void as well, since it documents the expected override contract for custom formatters.

gems/faraday/2.5/faraday.rbs
```diff
@@ -225,6 +225,12 @@ module Faraday
   end
 
   class Response
+    class Logger < Middleware
```
yykamei, line 228: `Faraday::Response::Logger` also defines `on_error(exc)` in the implementation, but it is missing from this signature.
Could you add it here?

```rbs
def on_error: (untyped exc) -> void
```

gems/faraday/2.5/faraday.rbs
```diff
@@ -225,6 +225,12 @@ module Faraday
...
+      def initialize: (untyped app, ?untyped logger, ?::Hash[Symbol, untyped] options) ?{ (Logging::Formatter formatter) -> void } -> void
+      def call: (Env env) -> untyped
+      def on_complete: (Env env) -> untyped
```
yykamei: Since `on_complete` only delegates to `@formatter.response(env)` and the formatter method is side-effect-oriented, I think this should be `void` rather than `untyped`.

```suggestion
      def on_complete: (Env env) -> void
```

gems/faraday/2.5/faraday.rbs
```diff
@@ -264,6 +270,42 @@ module Faraday
...
+      def error: (?untyped progname) ?{ () -> untyped } -> untyped
+      def fatal: (?untyped progname) ?{ () -> untyped } -> untyped
+
+      def request: (Env env) -> untyped
+      def response: (Env env) -> untyped
+      def filter: (untyped filter_word, untyped filter_replacement) -> ::Array[[untyped, untyped]]
+
+      private
+
+      def dump_headers: (untyped headers) -> String
```
yykamei, line 297: `dump_headers` can return `nil` when `headers` is `nil` in the implementation:

```ruby
return if headers.nil?
```

So this should be `String?` rather than `String`.

```suggestion
      def dump_headers: (untyped headers) -> String?
```

gems/faraday/2.5/faraday.rbs
```diff
@@ -264,6 +270,42 @@ module Faraday
...
+      def info: (?untyped progname) ?{ () -> untyped } -> untyped
+      def warn: (?untyped progname) ?{ () -> untyped } -> untyped
+      def error: (?untyped progname) ?{ () -> untyped } -> untyped
+      def fatal: (?untyped progname) ?{ () -> untyped } -> untyped
+
+      def request: (Env env) -> untyped
+      def response: (Env env) -> untyped
+      def filter: (untyped filter_word, untyped filter_replacement) -> ::Array[[untyped, untyped]]
+
+      private
```
yykamei, line 295: The implementation also has a private `log_errors?` method, which is used by `exception`, but it is missing from the RBS.

Could you add it alongside `log_headers?` and `log_body?`?

```rbs
def log_errors?: () -> boolish
```

yykamei (commented):
Thanks a lot for putting this together! Adding signatures for the logging formatter is really helpful for users writing custom formatters, and the example in `_test/` makes the intended usage very clear. Appreciate the contribution!

rossta:
@yykamei Thank you for the thorough review!

To be sure signatures were exposed in the correct VERSION/faraday.rbs, I reviewed Faraday sources for 2.5.x, 2.6.0, and 2.7.x releases.

Confirmed for 2.5/faraday.rbs:

- Faraday::Logging::Formatter, request, response, filter, delegated debug/info/warn/error/fatal, log_headers?, and log_body? are present from 2.5.0.
- Faraday::Response::Logger#on_complete is present from 2.5.0.
- log_errors?, on_error, and exception are not part of 2.5.
- dump_headers does not have the nil guard in 2.5, so I left it as String.

Patch availability for 2.7-only surface:

- Faraday::Response::Logger#on_error: added in 2.7.0.
- Faraday::Logging::Formatter#log_errors?: added in 2.7.0.
- Faraday::Logging::Formatter#error as the error callback: added in 2.7.0, replaced by exception in 2.7.2.
- Faraday::Logging::Formatter#exception: added in 2.7.2.
- dump_headers returning nil for nil headers: added in 2.7.8.

Implemented:

- Updated 2.5 side-effect methods to return void
- Added 2.7 overlay signatures for on_error, exception, and log_errors?
- Added 2.7 overloaded signature for dump_headers returning nil
- Added 2.7 formatter tests
- Amended commit

Verified:

```sh
bin/test gems/faraday/2.5
bin/test gems/faraday/2.7
```

Current amended commit is 2acd424. Please let me know whether this satisfies your comments.

gems/faraday/2.5/faraday.rbs
rossta: Added to `gems/faraday/2.7/faraday-2.7.rbs`

gems/faraday/2.5/faraday.rbs
rossta: Added signature overload in `gems/faraday/2.7/faraday-2.7.rbs`

gems/faraday/2.5/faraday.rbs
rossta: Done!

gems/faraday/2.5/faraday.rbs
rossta: Implemented in `gems/faraday/2.7/faraday-2.7.rbs`

gems/faraday/2.5/faraday.rbs
rossta: Implemented in `request`, `response`, `filter` changes in `gems/faraday/2.5`. Addressed `exception` in `gems/faraday/2.7/faraday-2.7.rbs`.

yykamei (commented):
APPROVE

PR #1041, opened 2026-06-17
activerecord: refine create/create! RBS signatures and update tests

kozy4324:
- Allow passing a block to `create` / `create!` .
- Add overloads for `attributes` as an Array of hashes and array return types.
- Add tests for `create` / `create!` .

## Note on file placement

The `create` / `create!` signatures are defined in `activerecord-8.0.rbs` rather than `activerecord.rbs`, because these methods are statically defined in Ruby source, not dynamically generated at runtime. The `activerecord.rbs` file already declares `extend ::ActiveRecord::Persistence::ClassMethods` on `ActiveRecord::Base`, so the type checker resolves them through that module naturally. Is this reasoning correct?


ksss (commented):
APPROVE

PR #1045, opened 2026-06-24
activerecord: Instance self type for the `:default` option argument in `belongs_to` 

felixefelip:
In Active Record `belongs_to` option `:default`, the self type in block is the instance type instead the singleton type. Doc: https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/belongs_to#:~:text=belongs%5Fto%20%3Aaccount%2C%20default%3A%20%2D%3E%20%7B%20company%2Eaccount%20%7D

In applications like Fizzy, it's a [very common use](https://github.com/basecamp/fizzy/blob/de2be807bb43b34cd7f60968af2e490666d1acca/app/models/card.rb#L6).

Without this fix, we have this output error:
```
activerecord-8.0.rb:37:36: [error] Type `singleton(::Test::Article)` does not have method `default_user`
│
└     belongs_to :user, default: -> { default_user }
                                      ~~~~~~~~~~~~

Detected 1 problem from 1 file

```

### DIFF
--- a/gems/activerecord/6.0/activerecord.rbs
+++ b/gems/activerecord/6.0/activerecord.rbs
@@ -42,9 +42,11 @@ module ActiveRecord
     extend ::ActiveRecord::Validations::ClassMethods
     extend ::ActiveSupport::Callbacks::ClassMethods
 
+    type belongs_to_default[T] = ^(T) [self: T] -> ::ActiveRecord::Base?
+
     def self.abstract_class=: (bool) -> void
     def self.scope: (Symbol, Proc) ?{ () -> untyped } -> void
-    def self.belongs_to: (Symbol, ?untyped, **untyped) -> void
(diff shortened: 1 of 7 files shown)


gems/activerecord/6.0/activerecord.rbs
```diff
@@ -42,9 +42,11 @@ module ActiveRecord
     extend ::ActiveRecord::Validations::ClassMethods
     extend ::ActiveSupport::Callbacks::ClassMethods
 
+    type belongs_to_default[T] = ^(T) [self: T] -> ::ActiveRecord::Base?
```
felixefelip, line 45: For the Active Record 6, I couldn't find a test setup, so I didn't add a test to this change. But if it's necessary, I'd be happy to add tests for this version as well.

ksss: I don't think we need to use a type alias.

ksss (commented):
I apologize for the delay in reviewing.

PR #1049, opened 2026-07-13
activerecord: Add filter keyword argument to in_order_of

dak2:
`in_order_of` accepts an optional `filter:` keyword, but the activerecord 8.0 signatures didn't allow it. 
Add `?filter: bool` to both definitions and a test case.

https://github.com/rails/rails/pull/51761
https://github.com/rails/rails/releases/tag/v8.0.0


ksss (commented):
APPROVE

PR #1050, opened 2026-07-14
Drop simplecov

sferik:
SimpleCov ships its own RBS signatures from version 1.0.1 on (`sig/simplecov.rbs`, included in the packaged gem). They cover the current public API, including the 1.0 configuration redesign.

The signatures in this collection cover the 0.22 API. Since the 1.0 release they are stale in ways that actively wrong. Keeping the entry is not harmless, because `rbs collection install` prefers this collection’s entry over the signatures bundled in the installed gem. Removing the entry is what lets SimpleCov 1.x users pick up the gem’s own signatures.

This follows the same path as 07e5c5eac885c1dc034a34a130bbc69ed973d579 (Drop aws-sdk-s3). Projects on SimpleCov 0.22 that want to keep the old signatures can pin an earlier revision of the collection in their `rbs_collection.yaml`.

I maintain SimpleCov and am happy to adjust this however you prefer.


ybiquitous (commented):
Wow, great! I've confirmed that the RBS file is published. 👍🏼 
https://github.com/simplecov-ruby/simplecov/blob/v1.0.1/sig/simplecov.rbs

APPROVE

ksss:
I have one additional point. (I have no intention of dissuading you from merging.)

The rbs collection searches for signatures in the following order:

1. sig/ included in the gem itself
2. gem_rbs_collection
3. stdlib/ included in the rbs repo

This is defined as follows:
https://github.com/ruby/rbs/blob/789759781724998cd5df11db709037fd907c206f/lib/rbs/collection/config.rb#L66-L72

If the search is not being performed in this order, there is a possibility that it is a bug in rbs.

sferik:
Sorry, I was mistaken about the ordering. Closing.

PR #1057, opened 2026-07-30
Sinatra: Make erb block param optional

bolshakov:
In Sinatra's templates the `erb` method block parameter is optional


euglena1215 (commented):
APPROVE

